PR #30, opened 2018-06-22
Refactor data/inference engine from package ml.dmlc.treelite to ml.dmlc.treelite4j

jamesliu:
Data interface is changed to the directory from ml.dmlc.treelite to ml.dmlc.treelite4j


jamesliu:
@hcho3 Please help to review it. Thanks.

runtime/java/treelite4j/src/main/java/ml/dmlc/treelite4j/InferenceEngine.java
```diff
@@ -0,0 +1,6 @@
+package ml.dmlc.treelite4j;
+
+public interface InferenceEngine {
```
hcho3: Would it be useful if the JNI runtime implements `InferenceEngine` interface?

PR #56, opened 2018-10-11
Make function declaration in header a prototype

horgh:
Without this, compiling with strict prototypes flags fails. e.g.:

```
/treelite/include/treelite/c_api_common.h:37:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
 TREELITE_DLL const char* TreeliteGetLastError();
```


hcho3:
Thanks!

PR #68, opened 2019-01-08
Handle categorical splits with empty left_categories

yskn67:
A segmentation fault occurs when left_categories size is 0.


hcho3:
Thanks!

PR #86, opened 2019-03-15
Fix build for 32-bit platform

### LINKED ISSUE #85: Build fail on 32bit platforms like Rasp3b
It seems size_t is the same type as uint32_t on RPi, and causing build fails. 

```
treelite/include/treelite/common.h:372:15: error: redefinition of ‘T treelite::common::TextToNumber(const string&) [with T = unsigned int; std::__cxx11::string = std::__cxx11::basic_string<char>]’
 inline size_t TextToNumber(const std::string& str) {
               ^~~~~~~~~~~~
/home/pi/workplace/neo-ai/neo-ai-dlr/3rdparty/treelite/include/treelite/common.h:357:17: note: ‘T treelite::common::TextToNumber(const string&) [with T = unsigned int; std::__cxx11::string = std::__cxx11::basic_string<char>]’ previously declared here
 inline uint32_t TextToNumber(const std::string& str) {
                 ^~~~~~~~~~~~

```

hcho3:
We shouldn't assume that `size_t` is always 64-bit.

Closes #85

### DIFF
--- a/include/treelite/common.h
+++ b/include/treelite/common.h
@@ -288,9 +288,9 @@ inline T TextToNumber(const std::string& str) {
                 || std::is_same<T, int>::value
                 || std::is_same<T, int8_t>::value
                 || std::is_same<T, uint32_t>::value
-                || std::is_same<T, size_t>::value,
+                || std::is_same<T, uint64_t>::value,
                 "unsupported data type for TextToNumber; use float, double, "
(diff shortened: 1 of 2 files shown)


hcho3:
@ziyu-guo Can you review?

ziyu-guo:
LGTM

PR #97, opened 2019-06-05
add method: Model.set_tree_limit

sperlingxx:


sperlingxx:
Hi @hcho3 , can you help to review this :)

python/setup.py
```diff
@@ -34,7 +34,7 @@ def has_ext_modules(self):
 LIB_PATH = LIBPATH['find_lib_path'](basename='treelite')
 RT_PATH = LIBPATH['find_lib_path'](basename='treelite_runtime')
 
-if (not LIB_PATH) or (not RT_PATH) or (not os.path.isdir('../build/runtime')):
```
hcho3: Why are we changing this line?

sperlingxx: Because I don't have `../build/runtime`, and I also setup python wrapper successfully.

hcho3: Can you check if you can import `treelite.runtime` in your Python shell?

sperlingxx: I have rebuilt TreeLite on a Docker image.
1.  mkdir build && cd build && cmake .. && make -j4
2. rm -rf runtime 
3. cd ../python && python setup.py (remove `os.path.isdir('../build/runtime')`) install

After above steps, I found no exceptions thrown when I ran `import treelite.runtime` in Python shell.

sperlingxx: I reverted it anyway.

Issue #100, opened 2019-06-19
TestLightGBMIntegration::test_sparse_categorical_model test fails with MSVC and Apple Clang


hcho3:
tests/python/test_lightgbm_integration.py::TestLightGBMIntegration::test_sparse_categorical_model test generates code with high-cardinality categorical splits. For such splits, Treelite currently generates massively long `if` conditions. This doesn't cause issue with GCC, but crashes Apple Clang and freezes MSVC.

TODO. Implement a more compact representation for categorical splits.

PR #101, opened 2019-06-19
[CI] Run tests on Windows

hcho3:
I ended up finding some issue when the unit tests were run on the Windows target.

* More exhaustive method to locate vcvarsall.bat
* Do not use GCC on Windows
* Do not produce empty array literals, as they are forbidden by MSVC
* Per #100, Skip test_sparse_categorical_model() test on Windows for now

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -253,7 +253,7 @@ jobs:
       Move-Item -Path .\whl_content\treelite-*.data\data\treelite\treelite_runtime.dll -Destination .\runtime\native\lib
       Remove-Item .\whl_content -Force -Recurse
       Set-Location -Path .\build
-      cmake .. -G"Visual Studio 14 2015 Win64"
+      cmake .. -G"Visual Studio 15 2017 Win64"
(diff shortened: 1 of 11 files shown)


hcho3:
There is a bug in `subprocess.Popen()` on Windows: https://bugs.python.org/issue37380. This bug was fixed in latest versions of 3.7 and 3.8 (dev). I will try to update Python in the test runner and see if crash is fixed.

hcho3:
Using latest Python doesn't make any difference. I'd have to conclude that Azure Pipelines doesn't let a process to write to a temporary directory under`C:\\Users\\VssAdministrator\\AppData\\Local\\Temp`.

hcho3:
Note to myself: the LETOR model takes inordinately long time to compile on Windows (with MSVC), taking more than an hour (the job gets killed by Azure Pipelines). The same model takes < 5 minutes on Linux however. I suspect it has to do with [this](https://github.com/Microsoft/WSL/issues/873#issuecomment-425272829)

PR #112, opened 2019-09-11
Add FailSafeCompiler, to guarantee fast compilation time for large models

hcho3:
`FailSafeCompiler` is a new code generator that produces a compact, simple code to run decision trees. The advantage is that the generated code compiles quickly.

Example: 1600-tree model trained on the Yahoo dataset.

Method | Time to `export_lib()` (*)
-----|------------ 
Current codegen (`ast_native`) | 169.09 s
New codegen (`failsafe`) | 5.40 s

(*) We set `nthread=1` to use only a single thread

So far, we relied on multithreading to keep compile time manageable. However, I've seen customers use models larger than 100k trees. Thus, we need a way to compile extremely large models in reasonable amount of time (under 900 seconds).

**Usage**
```python
import treelite
model = treelite.Model.load('./yahoo_ranker.model', 'xgboost')
model.export_lib('gcc', './ref.so', params={'parallel_comp': 1600, 'quantize': 1},
                 verbose=True, nthread=1)
```

**Limitations**

* Support only numerical splits (no categorical split)
* All splits must share the same comparison op
* Support only gradient boosted trees (no random forest)
* Most of compilation parameters do not apply: `annotate_in`, `quantize`, `parallel_comp`, `code_folding_req`.
* `created_shared()` and `export_lib()` will only use a single thread when compiling the generated C code. This is okay, since the generated code is so small.


hcho3:
Example code

**main.c**
```c
#include "header.h"

// nodes[]: stores nodes from all decision trees
const struct Node nodes[] = {
  { 0x800001BC, 0.90382498502731323, 1, 2 }, { 0x8000027D, 0.23374500870704651, 3, 4 },
  { 0x64, 0.97388499975204468, 5, 6 }, { 0x8000015E, 0.25422501564025879, 7, 8 },
  // ...
};
// nodes_row_ptr[]: marks bounaries between decision trees. The nodes belonging to 
//                  Tree [i] are found in nodes[nodes_row_ptr[i]:nodes_row_ptr[i+1]]
const int nodes_row_ptr[] = {
  0, 375, 756, 1141, 1552, 1927, 2314, 2693, 3080, 3455, 3854, 4245, 4634, 5035,
  5452, 5843, 6222, 6623, 7024, 7407, 7786, 8179, 8552, 8955, 9352, 9725, 10124,
  // ...
};

size_t get_num_output_group(void) {
  return 1;
}

size_t get_num_feature(void) {
  return 700;
}

static inline float pred_transform(float margin) {
  return margin;
}

float predict(union Entry* data, int pred_margin) {
  float sum = 0.0f;

  for (int tree_id = 0; tree_id < 1600; ++tree_id) {
    int nid = 0;
    const struct Node* tree = &nodes[nodes_row_ptr[tree_id]];
    while (tree[nid].cleft != -1) {
      const unsigned feature_id = tree[nid].sindex & ((1U << 31) - 1U);
      const unsigned char default_left = (tree[nid].sindex >> 31) != 0;
      if (data[feature_id].missing == -1) {
        nid = (default_left ? tree[nid].cleft : tree[nid].cright);
      } else {
        nid = (data[feature_id].fvalue < tree[nid].info.threshold
               ? tree[nid].cleft : tree[nid].cright);
      }
    }
    sum += tree[nid].info.leaf_value;
  }

  sum += (float)(0.5);
  if (!pred_margin) {
    return pred_transform(sum);
  } else {
    return sum;
  }

}
```

**header.h**
```c
#include <stdlib.h>
#include <string.h>
#include <math.h>
#include <stdint.h>

union Entry {
  int missing;
  float fvalue;
};

union NodeInfo {
  float leaf_value;
  float threshold;
};

struct Node {
  unsigned int sindex;
  union NodeInfo info;
  int cleft;
  int cright;
};

size_t get_num_output_group(void);
size_t get_num_feature(void);
float predict(union Entry* data, int pred_margin);
```

hcho3:
Invoking CI again to compute code coverage

PR #124, opened 2019-10-02
Fixed a missing return in CompareWithOp

canonizer:
- otherwise, when `common.h` is included, the compiler would issue a warning about a missing return statement


canonizer:
@teju85 FYI.

hcho3:
Thanks!

PR #127, opened 2019-10-08
Install fmtlib headers in CMake

### LINKED ISSUE #126: [BUG] treelite/common.h depends on fmt/format.h but does not install those headers
`treelite/common.h` seems to #include `fmt/format.h` and use it inside [here](https://github.com/dmlc/treelite/blob/master/include/treelite/common.h#L247). However, cmake install command does not include this header. Thus, if one is depending on cmake installed treelite, one will face a compilation error in their project due to unavailable format.h.

hcho3:
Closes #126.

@teju85 @canonizer @dantegd Please review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,3 +149,5 @@ endif()
 
 install(DIRECTORY include/treelite DESTINATION include
         FILES_MATCHING PATTERN "*.h")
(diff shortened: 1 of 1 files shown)


hcho3:
@teju85 Thanks for your review.

PR #129, opened 2019-10-18
Save libs relative to build folder

apivovarov:
This is needed for cases when we simultaneously build project for different architectures. For Example when building the project using Android Gradle.

Libs location now:
```
./build/lib/libtreelite4j.so
./build/lib/libtreelite_runtime.so
./build/lib/libtreelite.so
```


apivovarov:
Eventually all tests are passed except of "codecov/project". I think the failure has nothing to do with the patch itself.

PR #132, opened 2019-11-13
chage `format` to `model_format` in example

fnand:
The example is out of date and uses the argument `format` instead of `model_format` for Model.load as correctly specified in the documentation.


hcho3:
Thanks a lot!

PR #135, opened 2019-12-03
Allow users to disable explicitly OpenMP

mpuccio:
Hi,
for our project we need to have builds without OpenMP enabled. I am not sure if this is the smartest way of achieving this, however for us worked well...
Let me know if for some reason this might not be advisable.

Thanks a lot!


hcho3:
Thanks!

PR #138, opened 2020-01-13
make more model information available in runtime 

sperlingxx:
While doing inference work with binary generated by TreeLite, one thing really annoying is lack of meta information like the task of model (Regression?BinaryClassification?).  So, the aim of this PR is making more meta information of model available in TreeLite runtime. In terms of me, the most useful information exposed in PR is the name of `pred_transform`.
Could you help to review this PR? @hcho3


sperlingxx:
It is weird that get functions added in this PR can't be loaded in win32 test, while everything is all right in linux/osx tests.

hcho3:
I think you need to add the extra functions to https://github.com/dmlc/treelite/blob/9cfe9fe4e5bdc71cb8e17363afd0374980fc96bd/src/compiler/native/header_template.h

hcho3:
Thanks!

PR #141, opened 2020-02-03
[treelite4j] enforce rich model info on jvm runtime

sperlingxx:
This PR is to enable query of extra model information (which has been applied in #138 ) on Java predictor.


sperlingxx:
@hcho3 How can I fix the code coverage decreasion problem?

hcho3:
@sperlingxx Code coverage is currently broken. I will fix it later.

PR #144, opened 2020-02-11
Fix loading latest XGBoost binary model.

trivialfis:
For XGBoost >= 1.0.0, depends on https://github.com/dmlc/xgboost/pull/5281 .

### DIFF
--- a/src/frontend/xgboost.cc
+++ b/src/frontend/xgboost.cc
@@ -5,12 +5,11 @@
  * \author Philip Cho
  */
 
-#include <algorithm>
-#include <bits/stdint-uintn.h>
 #include <dmlc/data.h>
 #include <dmlc/memory_io.h>
 #include <treelite/frontend.h>
 #include <treelite/tree.h>
+#include <algorithm>
 #include <memory>
 #include <queue>
 #include <cstring>
--- a/src/frontend/xgboost.cc
+++ b/src/frontend/xgboost.cc
@@ -5,6 +5,7 @@
  * \author Philip Cho
  */
 
+#include <algorithm>
 #include <dmlc/data.h>
 #include <dmlc/memory_io.h>
 #include <treelite/frontend.h>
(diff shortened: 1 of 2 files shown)


trivialfis:
@hcho3 Could you please take a look and test it yourself with  https://github.com/dmlc/xgboost/pull/5281 when time allows?  I want to merge 2 PRs together, making sure that everything works as expected.

hcho3:
@trivialfis The Treelite CI is using XGBoost 0.90 for integration tests. We need to test 1.0.0. Should I try the models you gave me earlier?

trivialfis:
@hcho3 For loading 0.90 the CI tests should be sufficient.  Just testing treelite in combination with https://github.com/dmlc/xgboost/pull/5281 .  I tested myself, but it will be great that you can test it too.

trivialfis:
Running the test I added in treelite should be sufficient.

src/frontend/xgboost.cc
```diff
@@ -390,6 +399,17 @@ inline treelite::Model ParseStream(dmlc::Stream* fi) {
 
   // set global bias
   model.param.global_bias = static_cast<float>(mparam_.base_score);
+  std::vector<std::string> exponential_family {
+    "count:poisson", "reg:gamma", "reg:tweedie"
```
JohnZed: do we need to do something special for `survival:cox`? or assert that it is not supported?

tests/python/test_xgboost_integration.py
```diff
@@ -77,3 +77,28 @@ def test_xgb_iris(self):
       assert predictor.pred_transform == 'max_index'
       assert predictor.global_bias == 0.5
       assert predictor.sigmoid_alpha == 1.0
+
+  def test_logistic(self):
+    np.random.seed(0)
+    kRows = 16
```
JohnZed: Is there a similar test for the exponential families? You could parametrize and reuse this with just a different randint limit for count:poisson probably.

src/frontend/xgboost.cc
hcho3: No, it is not necessary, since the label is not in log scale for `survival:cox`.

tests/python/test_xgboost_integration.py
trivialfis: Adding it.

src/frontend/xgboost.cc
trivialfis: @hcho3 has better understanding of survival model than me.

trivialfis:
> I ran the XGB integration test on my machine successfully with dmlc/xgboost#5281.

Great!

trivialfis:
@hcho3 Shall I merge the PR in XGBoost ?

hcho3:
Yes

PR #145, opened 2020-02-14
[jvm-runtime] prepare for spark binding

sperlingxx:
The aim of this PR is making preparation for Spark binding development, which includes:
* custom serialization for java predictor
* scala predictor based on java one
* remove mutex lock on predictBatch method when prediction task is single-threaded


sperlingxx:
@hcho3 could you help to review this?

hcho3:
@sperlingxx Can you hold off any follow-up work after this? Within 2 weeks, I am refactoring the runtime to allow for multiple data types for the input matrix. As a result, the C API will also change too.

PR #151, opened 2020-03-20
Optional add fmt 3rdparty lib

mizinovmv:


hcho3:
No, fmtlib is essential in codegen. It cannot be made optional.

mizinovmv:
@hcho3 ok, how resolve error with many dependencies with fmt?

CMake Error at libs/<dependency>/libs/fmt/CMakeLists.txt:157 (add_library):
  add_library cannot create target "fmt" because another target with the same
  name already exists.  The existing target is a shared library created in
  source directory
  "github/project/libs/treelite/3rdparty/fmt/fmt".  See
  documentation for policy CMP0002 for more details.

hcho3:
@mizinovmv Are you using Treelite as a submodule to another application? Is your application using fmtlib?

mizinovmv:
@hcho3  yes, treelite as git submodule of project.
Project has many submodules and one of them has fmt dependency too.

project:
-- treelite -> fmt
-- lib2 -> fmt
-- lib3 -> fmt

fmt versions are compatible.

hcho3:
@mizinovmv I see now what you are doing here. Did you run install fmtlib into the system using `make install`? If so, we could use `find_package()` directive to use system fmtlib instead.

mizinovmv:
@hcho3 I dont install system fmtlib.

I want something like this in cmake without any system libs:
In root CMakeLists.txt
set(lib1_FMT ON CACHE BOOL "on fmt")
add_subdirectory(libs/lib1)
set(TREELITE_FMT OFF CACHE BOOL "on fmt")
add_subdirectory(libs/treelite)

I dont know now how make this with find_package :(

hcho3:
We can detect existence of `fmt` CMake target. If the target already exists, then Treelite can use that target again.

hcho3:
@mizinovmv See if the new change works for you.

mizinovmv:
hcho3 yes, resolve problem, thanks!

PR #197, opened 2020-09-08
Add RapidJSON as a dependency

hcho3:
It will be useful for adding support for XGBoost JSON models.

@wphicks Can you review?


wphicks (approved):
LGTM

PR #196, opened 2020-08-29
Add support for multiple data types (will break up into smaller pull requests)

hcho3:
Addresses #95 and #111.

Trying again, since #130 failed. This time, I made the `Model` class to be polymorphic. This way, the amount of pointer indirection is minimized.

Summary: `Model` is an opaque container that wraps the polymorphic handle `ModelImpl<ThresholdType, LeafOutputType>`. The handle in turn stores the list of trees `Tree<ThresholdType, LeafOutputType>`. To unbox the `Model` container and obtain `ModelImpl<ThresholdType, LeafOutputType>`, use `Model::Dispatch(<lambda expression>)`. 

Also, upgrade to C++14 to access the generic lambda feature, which proved to be very useful in the dispatching logic for the polymorphic `Model` class.

EDIT. I will break up this PR into smaller PRs, once I get the whole system working together correctly.

TODOs
- [x] Turn the Model and Tree classes into template classes
- [x] Revise the string templates so that correct data types are used in the generated C code
- [x] Rewrite the model builder class
- [x] Revise the zero-copy serializer
- [x] Create an abstract matrix class that supports multiple data types (float32, float64 for now).
- [x] Redesign the C runtime API, using the abstract matrix class.
- [x] Ensure accuracy of scikit-learn models. To achieve the best results, use float32 for the input matrix and float64 for the split thresholds and leaf outputs.
- [x] Revise the JVM runtime.

PR #198, opened 2020-09-10
Create template classes for model representations to support multiple data types (Part of #196)

hcho3:
Extracted from #196 for easier reviewing.
(The CI tests in this PR won't pass due to missing parts. See #196 to look up the test results.)

* The `ModelImpl` class (created in #201) becomes the template class `ModelImpl<ThresholdType, LeafOutputType>`.
* Implement template classes `ModelImpl<ThresholdType, LeafOutputType>` and `TreeImpl<ThresholdType, LeafOutputType>` that contain the details of the tree ensemble model. The template classes are parameterized by the types of the thresholds and leaf outputs. Currently the following combinations are allowed:

| Threshold type | Leaf output type |
|----|----|
| float32 | float32 |
| float32 | uint32 |
| float64 | float64 |
| float64 | uint32 |

* Revise the zero-copy serialization protocol, to prepend the type information (threshold_type, leaf_output_type) to the serialized types so that the recipient will choose the correct `ModelImpl<ThresholdType, LeafOutputType>` to deserialize to.

* A run-time type dispatching system using the enum type `TypeInfo`. Users are able to dispatch a correct version of `ModelImpl<ThresholdType, LeafOutputType>`  by specifying a pair of `TypeInfo` values. We also implement a set of convenient functions, such as `InferTypeInfoOf<T>` that converts the template arg `T` into `TypeInfo` enum.

@canonizer @levsnv @jakirkham @wphicks Can you review this PR?

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -145,9 +145,9 @@ class Tree {
                 || std::is_same<LeafOutputType, float>::value
                 || std::is_same<LeafOutputType, double>::value,
                 "LeafOutputType must be one of uint32_t, float32 or float64");
-  static_assert(!std::is_same<ThresholdType, float>::value
-                || !std::is_same<LeafOutputType, double>::value,
-                "LeafOutputType cannot be float64 when ThresholdType is float32");
+  static_assert(std::is_same<ThresholdType, LeafOutputType>::value
+                || std::is_same<LeafOutputType, uint32_t>::value,
+                "Unsupported combination of ThresholdType and LeafOutputType");
   static_assert((std::is_same<ThresholdType, float>::value && sizeof(Node) == 48)
                 || (std::is_same<ThresholdType, double>::value && sizeof(Node) == 56),
                 "Node size incorrect");
@@ -555,7 +555,7 @@ class ModelImpl : public Model {
   inline size_t GetNumTree() const override {
     return trees.size();
   }
-  inline void SetTreeLimit(size_t limit) override {
+  void SetTreeLimit(size_t limit) override {
     return trees.resize(limit);
   }
 
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -78,6 +78,7 @@ class ContiguousArray {
 };
 
 /*! \brief in-memory representation of a decision tree */
+template <typename ThresholdType, typename LeafOutputType>
 class Tree {
  public:
   /*! \brief tree node */
@@ -87,8 +88,8 @@ class Tree {
     inline void Init();
     /*! \brief store either leaf value or decision threshold */
     union Info {
-      tl_float leaf_value;  // for leaf nodes
-      tl_float threshold;   // for non-leaf nodes
+      LeafOutputType leaf_value;  // for leaf nodes
+      ThresholdType threshold;   // for non-leaf nodes
     };
     /*! \brief pointer to left and right children */
     int32_t cleft_, cright_;
@@ -137,7 +138,19 @@ class Tree {
   };
 
   static_assert(std::is_pod<Node>::value, "Node must be a POD type");
-  static_assert(sizeof(Node) == 48, "Node must be 48 bytes");
+  static_assert(std::is_same<ThresholdType, float>::value
+                || std::is_same<ThresholdType, double>::value,
+                "ThresholdType must be either float32 or float64");
+  static_assert(std::is_same<LeafOutputType, uint32_t>::value
+                || std::is_same<LeafOutputType, float>::value
+                || std::is_same<LeafOutputType, double>::value,
+                "LeafOutputType must be either uint32_t, float32 or float64");
+  static_assert(!std::is_same<ThresholdType, float>::value
+                || !std::is_same<LeafOutputType, double>::value,
+                "LeafOutputType cannot be float64 when ThresholdType is float32");
+  static_assert((std::is_same<ThresholdType, float>::value && sizeof(Node) == 48)
+                || (std::is_same<ThresholdType, double>::value && sizeof(Node) == 56),
+                "Node incorrect size");
 
   Tree() = default;
   ~Tree() = default;
@@ -146,6 +159,7 @@ class Tree {
   Tree(Tree&&) noexcept = default;
   Tree& operator=(Tree&&) noexcept = default;
 
+  inline const char* GetFormatStringForNode();
   inline void GetPyBuffer(std::vector<PyBufferFrame>* dest);
   inline void InitFromPyBuffer(std::vector<PyBufferFrame>::iterator begin,
                                std::vector<PyBufferFrame>::iterator end);
@@ -153,7 +167,7 @@ class Tree {
  private:
   // vector of nodes
   ContiguousArray<Node> nodes_;
-  ContiguousArray<tl_float> leaf_vector_;
+  ContiguousArray<LeafOutputType> leaf_vector_;
   ContiguousArray<size_t> leaf_vector_offset_;
   ContiguousArray<uint32_t> left_categories_;
   ContiguousArray<size_t> left_categories_offset_;
@@ -225,19 +239,19 @@ class Tree {
    * \brief get leaf value of the leaf node
    * \param nid ID of node being queried
    */
-  inline tl_float LeafValue(int nid) const {
+  inline LeafOutputType LeafValue(int nid) const {
     return (nodes_[nid].info_).leaf_value;
   }
   /*!
    * \brief get leaf vector of the leaf node; useful for multi-class random forest classifier
    * \param nid ID of node being queried
    */
-  inline std::vector<tl_float> LeafVector(int nid) const {
+  inline std::vector<LeafOutputType> LeafVector(int nid) const {
     if (nid > leaf_vector_offset_.Size()) {
       throw std::runtime_error("nid too large");
     }
-    return std::vector<tl_float>(&leaf_vector_[leaf_vector_offset_[nid]],
-                                 &leaf_vector_[leaf_vector_offset_[nid + 1]]);
+    return std::vector<LeafOutputType>(&leaf_vector_[leaf_vector_offset_[nid]],
+                                       &leaf_vector_[leaf_vector_offset_[nid + 1]]);
   }
   /*!
    * \brief tests whether the leaf node has a non-empty leaf vector
@@ -253,7 +267,7 @@ class Tree {
    * \brief get threshold of the node
    * \param nid ID of node being queried
    */
-  inline tl_float Threshold(int nid) const {
+  inline ThresholdType Threshold(int nid) const {
     return (nodes_[nid].info_).threshold;
   }
   /*!
@@ -346,7 +360,7 @@ class Tree {
    * \param cmp comparison operator to compare between feature value and
    *            threshold
    */
-  inline void SetNumericalSplit(int nid, unsigned split_index, tl_float threshold,
+  inline void SetNumericalSplit(int nid, unsigned split_index, ThresholdType threshold,
                                 bool default_left, Operator cmp);
   /*!
    * \brief create a categorical split
@@ -365,13 +379,13 @@ class Tree {
    * \param nid ID of node being updated
    * \param value leaf value
    */
-  inline void SetLeaf(int nid, tl_float value);
+  inline void SetLeaf(int nid, LeafOutputType value);
   /*!
    * \brief set the leaf vector of the node; useful for multi-class random forest classifier
    * \param nid ID of node being updated
    * \param leaf_vector leaf vector
    */
-  inline void SetLeafVector(int nid, const std::vector<tl_float>& leaf_vector);
+  inline void SetLeafVector(int nid, const std::vector<LeafOutputType>& leaf_vector);
   /*!
    * \brief set the hessian sum of the node
    * \param nid ID of node being updated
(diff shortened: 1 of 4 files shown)


include/treelite/tree.h
```diff
@@ -133,28 +135,40 @@ class Tree {
...
 
   static_assert(std::is_pod<Node>::value, "Node must be a POD type");
-  static_assert(sizeof(Node) == 48, "Node must be 48 bytes");
+  static_assert(std::is_same<ThresholdType, float>::value
+                || std::is_same<ThresholdType, double>::value,
+                "ThresholdType must be either float32 or float64");
+  static_assert(std::is_same<LeafOutputType, uint32_t>::value
+                || std::is_same<LeafOutputType, float>::value
+                || std::is_same<LeafOutputType, double>::value,
+                "LeafOutputType must be either uint32_t, float32 or float64");
```
wphicks:
```suggestion
                "LeafOutputType must be one of uint32_t, float32 or float64");
```

wphicks (changes requested):
Just one grammatical suggestion. Otherwise, everything looks great!

include/treelite/tree.h
hcho3: Thanks!

jakirkham:
There's a few things being passed as pointers. Would it make sense to pass them by reference instead? Or would there be issues with that?

include/treelite/tree_impl.h
```diff
@@ -348,6 +355,18 @@ inline void InitArrayFromPyBuffer(ContiguousArray<T>* vec, PyBufferFrame buffer)
   vec->UseForeignBuffer(buffer.buf, buffer.nitem);
 }
 
+inline void InitScalarFromPyBuffer(TypeInfo* scalar, PyBufferFrame buffer) {
```
jakirkham, line 358: Is it possible to pass this by reference to avoid a copy?

```suggestion
inline void InitScalarFromPyBuffer(TypeInfo* scalar, const PyBufferFrame& buffer) {
```

<br>

( I'm guessing there are other lines that would be updated if we change this as well. Though I'm guessing you know that much better than I 😄 )

hcho3: No deep copy is involved when we copy `PyBufferFrame`, since it holds memory references to existing objects. Do you think we should still avoid copying it?

hcho3:
@jakirkham 

> Would it make sense to pass them by reference instead? Or would there be issues with that?

Cpplint and Google C++ style guide warn against using non-const reference parameters in functions.

jakirkham:
Interesting, I'm curious why Google thinks a potentially null pointer is safer than a reference 😄

include/treelite/tree_impl.h
jakirkham: Thanks for the info. I don't have deep thoughts here. It was just something that occurred to me when looking at the code. It sounds like you know better here. Feel free to mark as resolved 🙂

hcho3:
@jakirkham The idea is to explicitly mark "output" parameters. `FooBar(&a)` clearly shows that `a` is modified, whereas `FooBar(a)` isn't clear whether `a` is being modified or not.

include/treelite/tree.h
```diff
@@ -133,28 +135,40 @@ class Tree {
...
   static_assert(std::is_pod<Node>::value, "Node must be a POD type");
-  static_assert(sizeof(Node) == 48, "Node must be 48 bytes");
+  static_assert(std::is_same<ThresholdType, float>::value
+                || std::is_same<ThresholdType, double>::value,
+                "ThresholdType must be either float32 or float64");
+  static_assert(std::is_same<LeafOutputType, uint32_t>::value
+                || std::is_same<LeafOutputType, float>::value
+                || std::is_same<LeafOutputType, double>::value,
+                "LeafOutputType must be one of uint32_t, float32 or float64");
+  static_assert(!std::is_same<ThresholdType, float>::value
```
canonizer: Any specific reason for this?

include/treelite/tree.h
```diff
@@ -133,28 +135,40 @@ class Tree {
...
+                || !std::is_same<LeafOutputType, double>::value,
+                "LeafOutputType cannot be float64 when ThresholdType is float32");
+  static_assert((std::is_same<ThresholdType, float>::value && sizeof(Node) == 48)
+                || (std::is_same<ThresholdType, double>::value && sizeof(Node) == 56),
+                "Node incorrect size");
```
canonizer: "Node size incorrect"

include/treelite/typeinfo.h
```diff
@@ -0,0 +1,163 @@
...
+    return "invalid";
+  case treelite::TypeInfo::kUInt32:
+    return "uint32";
+  case treelite::TypeInfo::kFloat32:
+    return "float32";
+  case treelite::TypeInfo::kFloat64:
+    return "float64";
+  default:
+    throw std::runtime_error("Unrecognized type");
+    return "";
```
canonizer: Is `return` required after `throw`?

include/treelite/typeinfo.h
```diff
@@ -0,0 +1,163 @@
...
+ * \tparam Dispatcher Function object that takes in one template arg.
+ *         It must have a Dispatch() static function.
+ * \tparam Parameter pack, to forward an arbitrary number of args to Dispatcher::Dispatch()
+ * \param type TypeInfo corresponding to the template arg T with which
+ *             Dispatcher<T>::Dispatch() is called.
+ * \param args Other extra parameters to pass to Dispatcher::Dispatch()
+ * \return Whatever that's returned by the dispatcher
+ */
+template <template<class> class Dispatcher, typename ...Args>
+inline auto DispatchWithTypeInfo(TypeInfo type, Args&& ...args) {
```
canonizer: It's hard to understand what the intended usage of `DispatchWithTypeInfo()` / `DispatchWithModelTypes()` is. Could you add an example (or reference a place in code) demonstrating it?

include/treelite/typeinfo.h
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+/*!
+ * \brief Convert a template type into a type info
+ * \tparam template type to be converted
+ * \return TypeInfo corresponding to the template type arg
+ */
+template <typename T>
+inline TypeInfo InferTypeInfoOf() {
```
canonizer: `TypeToInfo()` sounds like a slightly better name (but that's up to you).

include/treelite/typeinfo.h
```diff
@@ -0,0 +1,163 @@
...
+  if (std::is_same<T, uint32_t>::value) {
+    return TypeInfo::kUInt32;
+  } else if (std::is_same<T, float>::value) {
+    return TypeInfo::kFloat32;
+  } else if (std::is_same<T, double>::value) {
+    return TypeInfo::kFloat64;
+  } else {
+    throw std::runtime_error(std::string("Unrecognized Value type") + typeid(T).name());
```
canonizer: Is runtime type information already used in treelite?

include/treelite/tree.h
```diff
@@ -424,18 +471,25 @@ struct Model {
   ModelParam param;
```
canonizer, line 529: `num_feature`, `num_output_group`, `random_forest_flag` and `param` do not depend on template parameters; I think they would be better placed in the base class (`Model`).

include/treelite/tree.h
```diff
@@ -424,18 +471,25 @@ struct Model {
...
-  inline void InitFromPyBuffer(std::vector<PyBufferFrame> frames);
-  inline Model Clone() const;
+  ModelImpl() = default;
+  ~ModelImpl() = default;
+  ModelImpl(const ModelImpl&) = delete;
+  ModelImpl& operator=(const ModelImpl&) = delete;
+  ModelImpl(ModelImpl&&) noexcept = default;
+  ModelImpl& operator=(ModelImpl&&) noexcept = default;
+
+  inline ModelParam GetParam() const override;
```
canonizer: `GetParam()`, `GetNumFeature()`, `GetNumOutputGroup()` and `GetRandomForestFlag()` could be defined in the base class if the corresponding fields are moved to the base class.

include/treelite/tree.h
```diff
@@ -405,10 +419,43 @@ static_assert(std::is_standard_layout<ModelParam>::value,
 inline void InitParamAndCheck(ModelParam* param,
                               const std::vector<std::pair<std::string, std::string>>& cfg);
 
+class Model {
```
canonizer: Consider moving the refactoring of `struct Model` into `class Model` and `class ModelImpl` (without template parameters) into a separate pull request.

include/treelite/tree_impl.h
```diff
@@ -520,120 +514,141 @@ Tree::GetCategoricalFeatures() const {
   return result;
 }
 
+template <typename ThresholdType, typename LeafOutputType>
 inline int
-Tree::LeftChild(int nid) const {
+Tree<ThresholdType, LeafOutputType>::LeftChild(int nid) const {
```
canonizer: There are lots of small methods like these in `Tree`, `Model` and `ModelImpl`; I wonder if it would be better to have them inside the class definition. (I don't know the original reason behind having them outside of the class definition, so this is just a suggestion).

hcho3: The `tree.h` header was getting very long, so I moved all implementations to a separate file.

include/treelite/typeinfo.h
hcho3: Without it, some old compilers like GCC 5 complains about a missing return value.

include/treelite/typeinfo.h
hcho3: Do you mean the built-in RTTI? Yes, since we use `dynamic_cast<>` to downcast `Model*` into `ModelImpl<>*`.

include/treelite/typeinfo.h
hcho3: Here is the idea. Let's say we are given a `TypeInfo` variable `type` at runtime. We want to call `Dispatcher<T>::Dispatch()` method where the template arg `T` corresponds to `type`, and `Dispatcher<T>` is a class that contains a custom function that we'd like to run.

For a concrete example, see https://github.com/dmlc/treelite/pull/198/files#diff-409f38fcc1a201a94e70922bbf7c44a6R777-R788 and https://github.com/dmlc/treelite/pull/196/files#diff-65416f46630c396d5db3df0ac92a0a02R130-R148.

include/treelite/typeinfo.h
hcho3: Sounds good.

include/treelite/tree.h
hcho3: Will create a separate PR.

include/treelite/tree.h
hcho3: This is to restrict the combination of ThresholdType and LeafOutputType as follows:

Threshold type | Leaf output type
-- | --
float32 | float32
float32 | uint32
float64 | float64
float64 | uint32

These combinations cover the use case of LightGBM, XGBoost, and scikit-learn.

We can allow more combinations in the future, as needs arise.

include/treelite/tree.h
hcho3: Fixed.

include/treelite/tree.h
hcho3: Fixed in #201

include/treelite/tree.h
hcho3: Fixed in #201

include/treelite/tree.h
hcho3: #201 has been merged.

include/treelite/tree_impl.h
hcho3: Fixed in #201

include/treelite/typeinfo.h
hcho3: Changed to `TypeToInfo()`.

hcho3:
@canonizer Can I get another round of review for this PR?

include/treelite/tree.h
canonizer: Note that this assertion allows, e.g. `ThresholdType == double && LeafOutputType == float`, which is not among the supported combinations.

I think it's better to replace this assertion with `LeafOutputType == ThresholdType || LeafOutputType == uint32`.

include/treelite/tree.h
```diff
@@ -525,7 +555,7 @@ class ModelImpl : public Model {
   inline size_t GetNumTree() const override {
     return trees.size();
   }
-  void SetTreeLimit(size_t limit) override {
+  inline void SetTreeLimit(size_t limit) override {
```
canonizer: `inline` here is redundant, as a function defined inside a `class`/`struct` definition is implicitly inline.

include/treelite/tree.h
```diff
@@ -474,12 +488,25 @@ class Model {
...
   Model() = default;
   virtual ~Model() = default;
-  inline static std::unique_ptr<Model> Create();
   Model(const Model&) = delete;
   Model& operator=(const Model&) = delete;
   Model(Model&&) = default;
   Model& operator=(Model&&) = default;
 
+  template <typename ThresholdType, typename LeafOutputType>
+  inline static std::unique_ptr<Model> Create();
```
canonizer, line 497: Consider moving this function into `ModelImpl`, as it depends on `ThresholdType` and `LeafOutputType`.

include/treelite/tree.h
```diff
@@ -474,12 +488,25 @@ class Model {
...
+  inline static std::unique_ptr<Model> Create(TypeInfo threshold_type, TypeInfo leaf_output_type);
+  inline TypeInfo GetThresholdType() const {
+    return threshold_type_;
+  }
+  inline TypeInfo GetLeafOutputType() const {
+    return leaf_output_type_;
+  }
+  template <typename Func>
+  inline auto Dispatch(Func func);
```
canonizer, line 506: Are these `Dispatch()` functions used anywhere?

canonizer (approved):
Approved, provided that the remaining comments are addressed.

include/treelite/tree.h
hcho3: Yes, it's used in the C code generator (excluded from this PR). See #196 to search for actual uses.

include/treelite/tree.h
hcho3: Thanks for the suggestion. I will fix it.

include/treelite/tree.h
hcho3: It is actually not possible to move the function, since it needs access the base class's members `threshold_type_` and `leaf_output_type_`. The members were moved to the base class per your previous suggestion.

include/treelite/tree.h
hcho3: Fixed in the latest commit.

include/treelite/tree.h
hcho3: Fixed in the latest commit.

include/treelite/tree.h
hcho3: Here is an example: https://github.com/dmlc/treelite/pull/196/files#diff-e85efd30a12be68c2d66cb6a7b674d48R124

PR #201, opened 2020-09-14
Refactor struct Model -> class Model + class ModelImpl (Part of #196)

hcho3:
Extracted from #198, which was extracted from #196 

Per https://github.com/dmlc/treelite/pull/198#discussion_r487349962, I created a separate pull request that just consists of the refactoring of `struct Model` into `class Model` and `class ModelImpl`. I also moved small functions' definitions back to `tree.h` header.

List of changes:
* Upgrade C++ standard to C++14.
* Split `struct Model` into `class Model` and `class ModelImpl`. The `ModelImpl` class will soon become a template class in order to hold `Tree` objects with uint32, float32, or float64 type. The `Model` class will become an abstract class so as to avoid exposing `ModelImpl` to external interface. (It's very hard to pass template classes through a FFI boundary.)
* Change signature of methods that return `Model`, since `Model` is now an abstract class. These functions now return `std::unique_ptr<Model>`.
* Move bodies of tiny methods from `tree_impl.h` to `tree.h`. This will reduce verbosity once `ModelImpl` becomes a template class.

@canonizer


canonizer (approved):
A list of changes in the pull request description would be useful. Otherwise looks good.

PR #202, opened 2020-09-22
Allow parsing of XGBoost models from JSON

wphicks:
Provide methods to support parsing of models from XGBoost's new JSON format


src/frontend/xgboost_json.cc
```diff
@@ -5,17 +5,31 @@
  * \author Hyunsu Cho
```
hcho3, line 5: Make sure to add your name here.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +39,517 @@ namespace frontend {
...
+#else
+  FILE * fp = fopen(filename, "r");
+#endif
+
+  auto input_stream = std::make_unique<rapidjson::FileReadStream> (
+      fp,
+      readBuffer,
+      sizeof(readBuffer)
+  );
+  *out = std::move(ParseStream(std::move(input_stream)));
```
hcho3, line 58: Don't forget to run `std::fclose()` to the file pointer `fp`.

hcho3:
@wphicks Now the linter no longer complains about non-const reference parameter.

cmake/ExternalLibs.cmake
```diff
@@ -45,7 +45,7 @@ add_library(RapidJSON::rapidjson ALIAS rapidjson)
 
 # Google C++ tests
 if(BUILD_CPP_TEST)
-  find_package(GTest)
+  find_package(GTest REQUIRED CONFIG)
```
hcho3: We shouldn't put `REQUIRED` here, since we want to optionally fetch gtest from GitHub if it's not available in the system prefix. (See the following `if` statement.)

hcho3: Also, what does `CONFIG` do here? Is it necessary?

wphicks: Thanks! `REQUIRED` was a debugging artifact; I've just removed it. `CONFIG` switches us from [module mode](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature-and-module-mode) to [config mode](https://cmake.org/cmake/help/latest/command/find_package.html#full-signature-and-config-mode) for package discovery. Config mode is generally the more modern and preferred way to set up packages in CMake, but they still default to module mode to allow library developers to easily override default configs generated by CMake.

In the case of GTest, module mode discovery does not correctly bring in GMock, only GTest. Switching to config mode is the easiest way to properly add GMock to the project.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,358 @@ namespace frontend {
...
+          assign_value("num_nodes", atoi(str), m_output) ||
+          check_cur_key("size_leaf_vector") || check_cur_key("num_deleted"));
+}
+
+/******************************************************************************
+ * RegTreeHandler
+ * ***************************************************************************/
+bool RegTreeHandler::StartArray() {
+  // Key "categories" not documented in schema
+  // Key "split_type" not documented in schema
```
hcho3: These keys are reserved for the categorical split feature, which is being actively developed. They are not part of the stable 1.2.0 release.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,358 @@ namespace frontend {
...
+bool IgnoreHandler::StartObject() { return push_handler<IgnoreHandler>(); }
+bool IgnoreHandler::Key(const char *str, std::size_t length, bool copy) {
+  return true; }
+bool IgnoreHandler::StartArray() { return push_handler<IgnoreHandler>(); }
+
+/******************************************************************************
+ * TreeParamHandler
+ * ***************************************************************************/
+bool TreeParamHandler::String(const char *str, std::size_t length, bool copy) {
+  // Key "num_deleted" not documented in schema
```
hcho3: `num_deleted` is an obsolete field which has been deprecated and removed.

src/frontend/xgboost_json.cc
wphicks: Excellent! I'll update the comments to reflect that.

src/frontend/xgboost_json.cc
wphicks: Interestingly, it's still being output by xgboost as far as I can tell. I'll update the comment to mention that it's deprecated. Thanks!

tests/python/test_xgboost_integration.py
```diff
@@ -130,3 +130,94 @@ def test_nonlinear_objective(tmpdir, objective, max_label, global_bias, toolchai
...
+    dtrain = xgboost.DMatrix(X_train, label=y_train)
+    dtest = xgboost.DMatrix(X_test, label=y_test)
+    param = {'max_depth': 8, 'eta': 1, 'silent': 1, 'objective': 'reg:linear'}
+    num_round = 10
+    bst = xgboost.train(param, dtrain, num_boost_round=num_round,
+                        evals=[(dtrain, 'train'), (dtest, 'test')])
+
+    # Serialize xgboost model
+    model_bin_path = os.path.join(tmpdir, 'serialized.model')
+    bst.save_model(model_bin_path)
```
hcho3, line 157: This line is currently crashing in the CI: https://travis-ci.org/github/dmlc/treelite/jobs/731392804

wphicks: Interesting. I wasn't able to reproduce this locally, and I don't think I touched anything that would have interfered with xgboost's model dump capabilities, but I'll see what I can find tomorrow.

wphicks: I was able to reproduce this both on this PR and on mainline. I'll create an issue and try to add a fix in a separate PR before we land this one.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+  return push_key_handler<LearnerHandler, treelite::Model>("learner",
+                                                           m_output);
+}
+
+bool XGBoostModelHandler::EndObject(std::size_t memberCount) {
+  if (memberCount != 2) {
+    return false;
+  }
+  m_output.random_forest_flag = false;
+  if (version[0] >= 1) {
```
hcho3, line 338:
```suggestion
  // Before XGBoost 1.0.0, the global bias saved in model is a transformed value.  After
  // 1.0 it's the original value provided by user.
  if (version[0] >= 1) {
```
Let's add a comment to describe why we need to transform the margin here.

src/frontend/xgboost/xgboost.h
```diff
@@ -0,0 +1,24 @@
+/*!
+ * Copyright (c) 2020 by Contributors
+ * \file xgboost_json.h
+ * \brief Methods for loading XGBoost models
```
hcho3:
```suggestion
 * \file xgboost.h
 * \brief Helper functions for loading XGBoost models
```

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+}
+
+
+}  // namespace details
+}  // namespace treelite
+
+namespace {
+template<typename StreamType>
+treelite::Model ParseStream(std::unique_ptr<StreamType> input_stream) {
+  std::shared_ptr<treelite::details::DelegatedHandler> handler = \
```
hcho3:
```suggestion
  std::shared_ptr<treelite::details::DelegatedHandler> handler =
```
The backslash is not necessary here.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+    treelite::details::DelegatedHandler::create();
+  rapidjson::Reader reader;
+
+  rapidjson::ParseResult result = reader.Parse(*input_stream, *handler);
+  if (!result) {
+    LOG(ERROR) << "Parsing error " << result.Code() << " at offset "
+               << result.Offset();
+    throw std::runtime_error(
+        "Provided JSON could not be parsed as XGBoost model");
```
hcho3:
```suggestion
    LOG(FATAL) << "Provided JSON could not be parsed as XGBoost model";
```
Let's use the DMLC logger here to be consistent with other parts of Treelite. (Note. The `tree.h` header uses a standard exception because it cannot include the DMLC header.)

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+class DelegatedHandler
+    : public rapidjson::BaseReaderHandler<rapidjson::UTF8<>, DelegatedHandler>,
+      public Delegator {
+
+ public:
+  /*! \brief create DelegatedHandler with initial RootHandler on stack */
+  static std::shared_ptr<DelegatedHandler> create() {
+    struct make_shared_enabler : public DelegatedHandler {};
+
+    std::shared_ptr<DelegatedHandler> new_handler = \
```
hcho3:
```suggestion
    std::shared_ptr<DelegatedHandler> new_handler =
```
Backslash not needed here.

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+/*! \brief handler for XGBoostModel objects from XGBoost schema*/
+class XGBoostModelHandler : public OutputHandler<treelite::Model> {
+ public:
+  using OutputHandler<treelite::Model>::OutputHandler;
+  bool StartArray();
+  bool StartObject();
+  bool EndObject(std::size_t memberCount);
+
+ private:
+  std::vector<unsigned> version;
```
hcho3, line 315:
```suggestion
  std::vector<unsigned> m_version;
```
to be consistent with `m_output`

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+/*! \brief handler for Learner objects from XGBoost schema*/
+class LearnerHandler : public OutputHandler<treelite::Model> {
+ public:
+  using OutputHandler<treelite::Model>::OutputHandler;
+  bool StartArray();
+  bool StartObject();
+  bool EndObject(std::size_t memberCount);
+
+ private:
+  std::string objective;
```
hcho3, line 303:
```suggestion
  std::string m_objective;
```
to be consistent with `m_output`

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+  bool Key(const char *str, std::size_t length, bool copy);
+  bool EndObject(std::size_t memberCount);
+  bool StartArray();
+  bool EndArray(std::size_t elementCount);
+
+ private:
+  DelegatedHandler() : delegates{}, result{} {};
+
+  std::stack<std::shared_ptr<BaseHandler>> delegates;
+  treelite::Model result;
```
hcho3, line 375:
```suggestion
  std::stack<std::shared_ptr<BaseHandler>> m_delegates;
  treelite::Model m_result;
```
to be consistent with `m_output`

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+  std::vector<double> sum_hessian;
+  std::vector<double> base_weights;
+  std::vector<int> leaf_child_counts;
+  std::vector<int> left_children;
+  std::vector<int> right_children;
+  std::vector<int> parents;
+  std::vector<int> split_indices;
+  std::vector<double> split_conditions;
+  std::vector<bool> default_left;
+  int num_nodes;
```
hcho3, line 261:
```suggestion
  std::vector<double> m_loss_changes;
  std::vector<double> m_sum_hessian;
  std::vector<double> m_base_weights;
  std::vector<int> m_leaf_child_counts;
  std::vector<int> m_left_children;
  std::vector<int> m_right_children;
  std::vector<int> m_parents;
  std::vector<int> m_split_indices;
  std::vector<double> m_split_conditions;
  std::vector<bool> m_default_left;
  int m_num_nodes;
```
to be consistent with `m_output`

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+    delegates.push(new_delegate);
+  }
+  /*! \brief pop handler off of stack, returning parsing responsibility to
+   *         previous handler on stack
+   */
+  void pop_delegate() override {
+    delegates.pop();
+  }
+  treelite::Model get_result();
+  bool Null();
```
hcho3, line 357: Let's explicitly add `override` keyword for all functions of derived classes that override the base class's virtual functions. This ensures that we are not adding new methods inadvertently.

For example:
```cpp
bool Null() override;

File: src/frontend/xgboost/xgboost_json.h [hunk 17]
```diff
@@ -0,0 +1,376 @@
...
+
+/*! \brief handler for array of objects of given type*/
+template <typename ElemType, typename HandlerType = BaseHandler>
+class ArrayHandler : public OutputHandler<std::vector<ElemType>> {
+ public:
+  using OutputHandler<std::vector<ElemType>>::OutputHandler;
+  bool Bool(ElemType b) {
+    this->m_output.push_back(b);
+    return true;
+  }
```
hcho3: This is confusing, since it's not clear whether this function is overriding the `bool Bool(bool)` function from the base class `BaseHandler`. Is this a trick to conditionally override `bool Bool(bool)` only when `ElemType == bool`? I'd advocate for clarity and reorganize as follows:
```cpp
typename std::enable_if<std::is_same<ElemType, bool>::value, bool>::type
Bool(ElemType b) override {
  this->m_output.push_back(static_cast<ElemType>(i));
  return true;
}

typename std::enable_if<!std::is_same<ElemType, bool>::value, bool>::type 
Bool(ElemType b) {
  return false;
}
```

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+
+  bool Int(int i) { return store_int<int>(i); }
+  bool Uint(unsigned u) { return store_int<unsigned>(u); }
+  bool Int64(int64_t i) { return store_int<int64_t>(i); }
+  bool Uint64(uint64_t u) { return store_int<uint64_t>(u); }
+  bool Double(ElemType d) {
+    this->m_output.push_back(d);
+    return true;
+  }
```
hcho3: Same comment applies. I suggest:
```cpp
typename std::enable_if<std::is_same<ElemType, double>::value, bool>::type
bool Double(ElemType d) override {
  this->m_output.push_back(d);
  return true;
}

typename std::enable_if<!std::is_same<ElemType, double>::value, bool>::type
bool Double(ElemType d) {
  return false;
}
```

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+  template <typename ArgType, typename IntType = ElemType>
+  typename std::enable_if<!std::is_integral<IntType>::value, bool>::type
+  store_int(ArgType i) {
+    return false;
+  }
+
+  bool Int(int i) { return store_int<int>(i); }
+  bool Uint(unsigned u) { return store_int<unsigned>(u); }
+  bool Int64(int64_t i) { return store_int<int64_t>(i); }
+  bool Uint64(uint64_t u) { return store_int<uint64_t>(u); }
```
hcho3:
```suggestion
  bool Int(int i) override { return store_int<int>(i); }
  bool Uint(unsigned u) override { return store_int<unsigned>(u); }
  bool Int64(int64_t i) override { return store_int<int64_t>(i); }
  bool Uint64(uint64_t u) override { return store_int<uint64_t>(u); }
```

Likewise, use `override` keyword whenever a function of a derived class overrides a virtual function of the base class.

src/frontend/xgboost/xgboost_json.h
```diff
@@ -0,0 +1,376 @@
...
+ protected:
+  /* \brief the output value constructed or modified during parsing */
+  OutputType &m_output;
+};
+
+/*! \brief handler for array of objects of given type*/
+template <typename ElemType, typename HandlerType = BaseHandler>
+class ArrayHandler : public OutputHandler<std::vector<ElemType>> {
+ public:
+  using OutputHandler<std::vector<ElemType>>::OutputHandler;
```
hcho3, line 164: TIL: I can use the `using` keyword to inherit the base class's constructors into the derived class: https://en.cppreference.com/w/cpp/language/using_declaration#Inheriting_constructors. This is quite handy feature.

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+          assign_value("num_class", static_cast<int>(std::max(atoi(str), 1)),
+                       m_output.num_output_group) ||
+          assign_value("num_feature", atoi(str), m_output.num_feature));
+}
+
+/******************************************************************************
+ * LearnerHandler
+ * ***************************************************************************/
+bool LearnerHandler::StartArray() {
+  return push_key_handler<IgnoreHandler>("eval_metrics"); }
```
hcho3: `eval_metrics` key is nowhere in the model schema. What is this line for?

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
-  return str;
+bool GBTreeModelHandler::StartObject() {
+  return push_key_handler<IgnoreHandler>("gbtree_model_param");
 }
 
+/******************************************************************************
+ * GradientBoosterHandler
+ * ***************************************************************************/
+bool GradientBoosterHandler::Uint(unsigned u) {
+  return check_cur_key("num_boosting_round");
```
hcho3: The key `num_boosting_round` does not appear anywhere in the model schema. What is this line for?

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+}
+bool GradientBoosterHandler::String(const char *str,
+                                    std::size_t length,
+                                    bool copy) {
```
hcho3, line 231:
```suggestion
                                    bool copy) {
  if (!check_cur_key("name")) {
    return false;
  }
```
We should first assert that the key is equal to "name".

src/frontend/xgboost_json.cc
```diff
@@ -25,45 +43,363 @@ namespace frontend {
...
+
+bool LearnerHandler::StartObject() {
+  // "attributes" key is not documented in schema
+  return (push_key_handler<LearnerParamHandler, treelite::Model>(
+              "learner_model_param", m_output) ||
+          push_key_handler<GradientBoosterHandler, treelite::Model>(
+              "gradient_booster", m_output) ||
+          push_key_handler<ObjectiveHandler, std::string>("objective",
+                                                          objective) ||
+          push_key_handler<IgnoreHandler>("attributes"));
```
hcho3, line 294: The `attributes` key appears nowhere in the current JSON model schema. What is this line for?

hcho3 (changes requested):
I like the overall design. The delegator pattern is super cool and I will make sure to learn it myself.

Some minor comments follow:

tests/cpp/test_frontend.cc
```diff
@@ -5,16 +5,498 @@
...
+  auto input_stream = rapidjson::MemoryStream(json_str.c_str(),
+                                              json_str.size());
+  std::shared_ptr<MockDelegator> delegator =
+    std::make_shared<MockDelegator>();
+
+  Tree output;
+  details::RegTreeHandler wrapped_handler {delegator, output};
+  MockObjectStarter handler {delegator, wrapped_handler};
+
+  EXPECT_CALL(*delegator, push_delegate).Times(1);
```
hcho3: Interleaving `EXPECT_CALL()` with calls to mock functions causes an undefined behavior: https://github.com/google/googletest/issues/2828. Please move all `EXPECT_CALL` to the beginning of this unit test.

tests/cpp/test_frontend.cc
```diff
@@ -5,16 +5,498 @@
...
+  std::shared_ptr<MockDelegator> delegator =
+    std::make_shared<MockDelegator>();
+
+  Model output;
+  details::LearnerParamHandler wrapped_handler {delegator, output};
+  MockObjectStarter handler {delegator, wrapped_handler};
+
+  rapidjson::Reader reader;
+  auto result = reader.Parse(input_stream, handler);
+}
```
hcho3: This test does not have any assertion. Is it sufficient that the test runs without crashing?

tests/cpp/test_frontend.cc
```diff
@@ -5,16 +5,498 @@
...
+    std::make_shared<MockDelegator>();
+
+  Model output;
+  details::LearnerParamHandler wrapped_handler {delegator, output};
+  MockObjectStarter handler {delegator, wrapped_handler};
 
-  Model model;
-  frontend::LoadXGBoostJSONModel(json_str, &model);
+  rapidjson::Reader reader;
+  auto result = reader.Parse(input_stream, handler);
```
hcho3: This test does not have any assertion. Is it sufficient that the test runs without crashing?

hcho3:
gMock is also quite nice. I will use it in the future.

tests/cpp/test_frontend.cc
wphicks: Yes, that's right. Because this is at such a high level in the tree, there are very few non-trivial checks that one could make for correctness here without invoking a larger stack of delegated handlers and thereby turning this into an integration test rather than a unit test. Since the Python test exercises the full stack, this unit test is designed solely to ensure that the C++ code does not immediately die at this level of parsing.

That being said, there's no reason to store the result here (unused variable). I'll pull that out.

tests/cpp/test_frontend.cc
wphicks: Same story as above. This one is even higher up the JSON tree, so it should be fully exercised by the integration test and has no non-trivial tests that can be performed in isolation.

tests/cpp/test_frontend.cc
wphicks: Ooh! Good catch. I'll fix that up right away.

src/frontend/xgboost_json.cc
wphicks: It does appear in the JSON output from XGBoost currently. Removing it prevents parsing of current XGBoost JSON models.

wphicks: Oh, and I note that it's not part of the schema at line 290. I'm hoping that's enough to clue future developers in to what's going on there, but we can certainly make that comment louder if necessary.

src/frontend/xgboost_json.cc
wphicks: Removed, thanks!

src/frontend/xgboost_json.cc
wphicks: Removed, thanks!

tests/cpp/test_frontend.cc
wphicks: Just added an explanatory comment to this and the other test without assertions. Also pulled out the unused variable from both.

wphicks:
Thanks so much for the review, @hcho3! Several good catches there. I've accepted all your suggested changes and resolved all other conversations except for the `attributes` key comment. That one is necessary for correct behavior right now, but please say the word if you think it needs more explanatory comments.

src/frontend/xgboost_json.cc
hcho3: Got it. Let me file a pull request to add `attributes` to the schema.

wphicks: Great! Thanks, @hcho3. Feel free to tag me for review.

hcho3:
@wphicks Can you also address my comments about the use of `override` keyword?

src/frontend/xgboost/xgboost_json.h
hcho3: @wphicks Please add `override` keyword to methods.

src/frontend/xgboost/xgboost_json.h
hcho3: ^^ @wphicks

wphicks:
Oh whoops! The Github UI wasn't showing some comments and I assumed they were outdated. Let me address all of those now. Sorry about that!

src/frontend/xgboost/xgboost_json.h
wphicks: I used the `m_` prefix on `m_output` just to distinguish it from the `output` argument. Unless we're using `m_` consistently throughout the rest of the codebase, I'd rather remove it from `m_output` and use something else, since I'm generally not a fan of polluting the variable name with an indication that it's a member variable if that's clear from the rest of the code. I'll do a quick check for how we're doing this elsewhere, and either match the rest of the codebase or rename `m_output`.

src/frontend/xgboost/xgboost_json.h
wphicks: Yes, that's right. Only `bool Bool(bool)` will ever be instantiated from the template, not `bool Bool(AnyOtherType)`. I'll go ahead and make it more explicit, though, as you suggest.

src/frontend/xgboost/xgboost_json.h
hcho3: @wphicks Got it. Then let's rename `m_output`.

src/frontend/xgboost/xgboost_json.h
wphicks: Ah, I'm beginning to remember why I went this direction to begin with. As written, the suggested change won't compile, but I'm assuming you meant something more along the lines of:

```c++
template <ArgType = ElemType>
typename std::enable_if<std::is_same<ArgType, bool>::value, bool>::type
Bool(ArgType b) override {
  this->m_output.push_back(b);
  return true;
}

template <ArgType = ElemType>
typename std::enable_if<std::is_same<ArgType, bool>::value, bool>::type
Bool(ArgType b) {
  return false;
}
```
This won't work either because a template cannot override a virtual function, so the base `Bool` is always called. The alternative is to create a `store_bool` function similar to what I did [here](https://github.com/dmlc/treelite/pull/202/files#diff-38fd03e89a1597a342c7600bd8ccede4R191-R207).

I guess I'd lean slightly toward leaving it as is and adding a comment than adding 2 additional methods, a layer of indirection, and more template metaprogramming, but I'm very happy to be overruled. It makes sense to me because I wrote it, but it's far more important that it makes sense to you and any other developers who will have to work with this later on.

What do you think, @hcho3?

src/frontend/xgboost/xgboost_json.h
wphicks: This one actually isn't an override.  `Null` and its ilk are not virtual in `BaseReaderHandler`. I think I got all the other overrides flagged in my last commit, though.

src/frontend/xgboost/xgboost_json.h
hcho3: Got it. In that case, please add a comment that clarifies the intention of `bool Bool(ElemType b)`.

src/frontend/xgboost/xgboost_json.h
hcho3: Interesting. I did not know that

hcho3:
Merged. Thanks a lot for your valuable contribution!

PR #206, opened 2020-09-25
[CI] Revise cpplint rule to allow non-const reference parameter

### LINKED ISSUE #205: Update linting rules to allow non-const reference parameters
Opening this issue to facilitate discussion around the prohibition of non-const reference parameters by the current linting infrastructure. This rule is [no longer part of the Google style guide](https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs), and there has been discussion about updating cpplint accordingly [here](https://github.com/google/styleguide/issues/563). Until that update happens, it might be useful to turn off that rule in our linting infrastructure.

The original intent of that rule was to clearly signal the intent of output parameters by forcing them to be pointers. This justification has a few issues:

1. As was discussed in the review for [this pull request](https://github.com/dmlc/treelite/pull/198), pointer output parameters open the possibility of passing null pointers, which requires additional safety checks. This is precisely the justification that the Google style guide now provides for encouraging reference parameters _instead_ of pointer output parameters.
2. There are some cases in which passing pointers as output parameters can increase the complexity of templated code, since pointers cannot take advantage of reference collapsing rules. This is the case for [this pull request](https://github.com/dmlc/treelite/pull/202/files#diff-1eac9db2992dcf96b1d414350dc4aa9dR155), where an attempt to convert the output parameter to a reference required an additional ~60 lines of code to handle additional safety checks as well as special cases that reference collapsing takes care of automatically.
3. As argued [here](https://eyakubovich.github.io/2018-11-27-google-cpp-style-guide-is-no-good/), using a pointer type to signal that a parameter is modified is something of an abuse of the semantic intent of C++ language features. Const provides mutability guarantees and signals the semantic intent that a parameter will not be modified, while passing a pointer carries no further semantic meaning than that the address of an object is being used in some way.
4. Using pointers as output parameters puts us at odds with C++ core guidelines, which recommends passing raw pointers only when the function specifically manipulates the address (taking advantage of the "pointeriness" of the parameter) and passing smart pointers only to take advantage of data ownership semantics.

What do other people think? Is there some reason for maintaining this rule that I'm missing? And/or some argument for doing inline suppression of this rule rather than disabling it project-wide?

hcho3:
Fixes #205

@wphicks Please review.

### DIFF
--- a/tests/ci_build/lint.py
+++ b/tests/ci_build/lint.py
@@ -1,0 +1,209 @@
+#!/usr/bin/env python3
+# pylint: disable=protected-access, unused-variable, locally-disabled, len-as-condition
+"""Lint helper to generate lint summary of source.
+
+Copyright by Contributors
+"""
+from __future__ import print_function
+import argparse
+import codecs
+import sys
+import re
+import os
+import cpplint
+from cpplint import _cpplint_state
+from pylint import epylint
+
+CXX_SUFFIX = set(['cc', 'c', 'cpp', 'h', 'cu', 'hpp'])
+PYTHON_SUFFIX = set(['py'])
+
+def filepath_enumerate(paths):
+    """Enumerate the file paths of all subfiles of the list of paths"""
+    out = []
+    for path in paths:
+        if os.path.isfile(path):
+            out.append(path)
+        else:
+            for root, dirs, files in os.walk(path):
+                for name in files:
+                    out.append(os.path.normpath(os.path.join(root, name)))
+    return out
+
+# pylint: disable=useless-object-inheritance
+class LintHelper(object):
+    """Class to help runing the lint and records summary"""
+
+    @staticmethod
+    def _print_summary_map(strm, result_map, ftype):
+        """Print summary of certain result map."""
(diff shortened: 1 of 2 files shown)


tests/ci_build/lint.py
```diff
@@ -0,0 +1,209 @@
...
+                            '--disable=' + ','.join(pylint_disable)]
+
+        self.pylint_cats = set(['error', 'warning', 'convention', 'refactor'])
+        # setup cpp lint
+        cpplint_args = ['.', '--extensions=' + (','.join(CXX_SUFFIX))]
+        _ = cpplint.ParseArguments(cpplint_args)
+        cpplint._SetFilters(','.join(['-build/c++11',
+                                      '-build/namespaces',
+                                      '-build/include',
+                                      '-runtime/references',
```
hcho3, line 69: I copied `lint.py` from dmlc-core and added this line to allow non-const reference parameter.

wphicks (approved):
Beautiful! I like that we're no longer wget-ing the configuration as well. LGTM!

PR #216, opened 2020-10-26
Support sklearn's extremely randomized trees

hcho3:
Add support for `ExtraTreesRegressor` and `ExtraTreesClassifier`. Since they use the same decision tree class as `RandomForestRegressor` and `RandomForestClassifier`, we can reuse most of the importing logic in `treelite.sklearn`.

cc @wphicks @canonizer


wphicks (approved):
LGTM

PR #217, opened 2020-10-29
Factor out common logics from xgboost and xgboost_json frontends

hcho3:
Factor out common functions `TransformGlobalBiasToMargin` and `SetPredTransform` from `xgboost` and `xgboost_json` frontends. This way, it's easier to add support for more objectives from XGBoost.

Depends on #218
Will rebase after #218 is merged to the mainline.

### DIFF
--- a/src/frontend/xgboost/xgboost.h
+++ b/src/frontend/xgboost/xgboost.h
@@ -7,10 +7,17 @@
 #ifndef TREELITE_FRONTEND_XGBOOST_XGBOOST_H_
 #define TREELITE_FRONTEND_XGBOOST_XGBOOST_H_
 
+#include <string>
+#include <vector>
 #include <cmath>
 
 namespace treelite {
+
+struct ModelParam;  // forward declaration
+
 namespace details {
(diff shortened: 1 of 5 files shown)


src/frontend/xgboost/xgboost.h
```diff
@@ -7,10 +7,17 @@
...
+#include <string>
+#include <vector>
 #include <cmath>
 
 namespace treelite {
+
+struct ModelParam;  // forward declaration
+
 namespace details {
+namespace xgboost {
```
wphicks, line 19: `ProbToMargin` seems like a struct that is generally useful (beyond just xgboost). Do we want it in the `xgboost` namespace?

wphicks (commented):
Just had one question, which I don't have strong feelings about. Everything else looks great!

src/frontend/xgboost/xgboost.h
hcho3: Only XGBoost requires ProbToMargin conversion on the global bias field.

PR #218, opened 2020-10-30
Bugfix in XGBoost-JSON frontend: Use correct field to set leaf outputs

hcho3:
In the XGBoost JSON model spec, it is the `split_conditions` field that holds the leaf output values, not `base_weights`.

Also expand test coverage for XGBoost-JSON, so that we don't run into similar bugs later.


wphicks (approved):
LGTM!

PR #229, opened 2020-12-04
New model specification

### LINKED ISSUE #133: LightGBM RF classifier with num_class > 2 is currently not supported
Hi sir,
I have a multi-class classification model generated from lightgbm, but I can not use treelite to export its. Here goes the error:
```
Check failed: !average_output_ Ill-formed LightGBM model file: cannot use random forest mode for multi-class classification
```
I only have such trained model, and I donot have the train dataset. So is there any method to load multi-class models directly  instead of retraining a model by ModelBuilder of treelite ?

### LINKED ISSUE #228: [RFC] New model specification
This is necessitated by #133. Also, we'd like to support integer leaf outputs.

## Motivation

Be precise and thorough in describing the kinds of tree models we would like to support in FIL and Treelite. Also address the shortcomings of the current model specification, as follows:
* The current spec fails to accommodate LightGBM’s multi-class classifiers where grove_per_class=True and average_tree_output=False (see the next section for definitions)
* Class outputs from cuML RF multi-class classifiers are currently encoded as one-hot-encoded leaf vectors, wasting memory.
The model spec does not align with the terminologies used in the FIL codebase.

## The Five Knobs
* **Output_type** (float/int): the type of output from each leaf node.
* **Grove_per_class** (True/False): whether we designate a subset of the trees to compute the prediction for each class. If True, the prediction for the i-th class is determined by the trees whose index is congruent to [i] modulo [num_class]. Only applicable if we are performing classification task with num_class > 2.
* **Num_class** (>= 1): the number of classes in the target label. Num_class should be >1 only when we’re performing multi-class classification. Otherwise, set num_class=1 (binary classification, regression etc).
* **Leaf_vector_size** (>= 1): Dimension of the output from each leaf node. If >1, each leaf produces a 1D vector output. If =1, each leaf produces a single scalar.
* **Average_tree_output** (True/False): After outputs from trees are combined, whether to divide the combined sum by the number of trees.

Since parameters **output_type**, **grove_per_class**, **num_class**, and **leaf_vector_size** are dependent on each other, we define an enumeration type **task_type** to capture the dependent relationship:
* **MULTI_CLF_GROVE_PER_CLASS**: implies output_type=float, grove_per_class=True, num_class > 1, leaf_vector_size=1.
* **MULTI_CLF_PROB_DIST_LEAF**: implies output_type=float, grove_per_class=False, num_class > 1, leaf_vector_size=num_class.
* **MULTI_CLF_CATEG_LEAF**: implies output_type=int, grove_per_class=False, num_class > 1, leaf_vector_size=1.
* **BINARY_CLF_REGR**: implies output_type=float, grove_per_class=False, num_class=1, leaf_vector_size=1.

## Full Catalog

Screen Shot 2020-10-30 at 1 30 28 PM

Note: we may in the future decide to support multi-output regression; in that case, we’d need to support the combination grove_per_class=False, num_class=1, leaf_vector_size>1. But we won’t support it now.

hcho3:
Implement #228 

Closes #133 
Closes #228

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -78,18 +78,102 @@ class ContiguousArray {
   bool owned_buffer_;
 };
 
+/*!
+ * \brief Enum type representing the task type.
+ *
+ * The task type places constraints on the parameters of TaskParameter. See the docstring for each
+ * enum constants for more details.
+ */
 enum class TaskType : uint8_t {
-  kBinaryClfRegr = 0,          // Binary classifier, regressor
-  kMultiClfGrovePerClass = 1,  // XGBoost, LightGBM multi-class classifier
-  kMultiClfProbDistLeaf = 2,   // Scikit-learn RF multi-class classifier
-  kMultiClfCategLeaf = 3       // cuML RF multi-class classifier
+  /*!
+   * \brief Catch-all task type encoding all tasks that are not multi-class classification, such as
+   *        binary classification, regression, and learning-to-rank.
+   *
+   * The kBinaryClfRegr task type implies the following constraints on the task parameters:
+   * output_type=float, grove_per_class=false, num_class=1, leaf_vector_size=1.
+   */
+  kBinaryClfRegr = 0,
+  /*!
+   * \brief The multi-class classification task, in which the prediction for each class is given
+   *        by the sum of outputs from a subset of the trees. We refer to this method as
+   *        "grove-per-class".
+   *
+   * In this setting, each leaf node in a tree produces a single scalar output. To obtain
+   * predictions for each class, we divide the trees into multiple groups ("groves") and then
+   * compute the sum of outputs of the trees in each group. The prediction for the i-th class is
+   * given by the sum of the outputs of the trees whose index is congruent to [i] modulo
+   * [num_class].
+   *
+   * Examples of "grove-per-class" classifier are found in XGBoost, LightGBM, and
+   * GradientBoostingClassifier of scikit-learn.
+   *
+   * The kMultiClfGrovePerClass task type implies the following constraints on the task parameters:
+   * output_type=float, grove_per_class=true, num_class>1, leaf_vector_size=1. In addition, we
+   * require that the number of trees is evenly divisible by [num_class].
+   */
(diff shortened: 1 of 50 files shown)


include/treelite/tree.h
```diff
@@ -77,6 +78,23 @@ class ContiguousArray {
   bool owned_buffer_;
 };
 
+enum class TaskType : uint8_t {
+  kBinaryClfRegr = 0,          // Binary classifier, regressor
```
canonizer: It would be better to describe the properties of the models specified by each of the constant, in addition to examples of frameworks that generate those models. For example, grove-per-class models distribute trees among classes in a round-robin fashion, and sum up predictions between trees of the same class only. Categorical leaf models store the class label in the tree leaves, and so on.

include/treelite/tree.h
```diff
@@ -519,12 +537,12 @@ class Model {
...
    */
   int num_feature;
-  /*! \brief number of output groups -- for multi-class classification
-   *  Set to 1 for everything else */
-  int num_output_group;
-  /*! \brief flag for random forest;
-   *  True for random forests and False for gradient boosted trees */
-  bool random_forest_flag;
+  /*! \brief Task type */
+  TaskType task_type;
```
canonizer, line 625: Any reason why `task_type` and `average_tree_output` are not part of `task_parameter`?

canonizer (approved):

include/treelite/tree.h
hcho3: The parameters in `task_parameter` are dependent on the choices of `task_type` and `average_tree_output`. See the table in #228.

include/treelite/tree.h
hcho3: I agree that we should clearly document the meaning of each parameter. Do the descriptions in #228 look reasonable to you? If so, I will add the description of each parameter as a docstring.

canonizer: The description for grove-per-class in the "five knobs" section looks reasonable, especially the part describing how the trees are assigned to classes. It would be good to have similar descriptions for other constants.

hcho3: Added detailed description in the docstring.

PR #231, opened 2020-12-05
Enhance error handling in XGBoost-JSON frontend: show error message + diagnosis

hcho3:
Improve error diagnosis by showing a human-readable message as well as the location of the error.

Before:
```
[23:27:30] ../src/frontend/xgboost_json.cc:374: Parsing error 16 at offset 438
treelite.util.TreeliteError: [23:27:30] ../src/frontend/xgboost_json.cc:376: Provided JSON could not be parsed
as XGBoost model
```

After:
```
treelite.util.TreeliteError: [23:25:20] ../src/frontend/xgboost_json.cc:450: Provided JSON could not be parsed
as XGBoost model. Parsing error at offset 438: Terminate parsing due to Handler error.
,1.0290909E1],"categories":[],"categories_nodes":[],"categories_segments":[],"categories_sizes":[],"
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/frontend/xgboost_json.cc
+++ b/src/frontend/xgboost_json.cc
@@ -128,8 +128,22 @@ std::unique_ptr<treelite::Model> LoadXGBoostJSONModel(const char* filename) {
 
 std::unique_ptr<treelite::Model> LoadXGBoostJSONModelString(const char* json_str, size_t length) {
   auto input_stream = std::make_unique<rapidjson::MemoryStream>(json_str, length);
-  auto error_handler = [](size_t offset) -> std::string {
-    return "";
+  auto error_handler = [json_str](size_t offset) -> std::string {
+    size_t cur = (offset >= 50 ? (offset - 50) : 0);
+    std::ostringstream oss, oss2;
+    for (int i = 0; i < 100; ++i) {
+      if (!json_str[cur]) {
+        break;
+      }
+      oss << json_str[cur];
+      if (cur == offset) {
+        oss2 << "^";
+      } else {
+        oss2 << "~";
+      }
+      ++cur;
+    }
+    return oss.str() + "\n" + oss2.str();
   };
   return ParseStream(std::move(input_stream), error_handler);
 }
--- a/src/frontend/xgboost_json.cc
+++ b/src/frontend/xgboost_json.cc
@@ -34,63 +34,6 @@ template <typename StreamType, typename ErrorHandlerFunc>
 std::unique_ptr<treelite::Model> ParseStream(std::unique_ptr<StreamType> input_stream,
                                              ErrorHandlerFunc error_handler);
 
-inline FILE* OpenFile(const char* filename) {
-#ifdef _WIN32
-  FILE* fp = std::fopen(filename, "rb");
-#else
-  FILE* fp = std::fopen(filename, "r");
-#endif
-  CHECK(fp) << "Failed to open file '" << filename << "': " << std::strerror(errno);
-  return fp;
-}
-
-class OwningFileReadStream {
- public:
-  explicit OwningFileReadStream(const char* filename)
-      : fp_(OpenFile(filename)), file_stream_(nullptr), read_buffer_("") {
-    file_stream_ = std::make_unique<rapidjson::FileReadStream>(
-        fp_, read_buffer_, sizeof(read_buffer_));
-  }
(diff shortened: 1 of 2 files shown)


src/frontend/xgboost_json.cc
```diff
@@ -23,16 +23,73 @@
...
+#ifdef _WIN32
+  FILE* fp = std::fopen(filename, "rb");
+#else
+  FILE* fp = std::fopen(filename, "r");
+#endif
+  CHECK(fp) << "Failed to open file '" << filename << "': " << std::strerror(errno);
+  return fp;
+}
+
+class OwningFileReadStream {
```
wphicks: I'm not sure I'm fully understanding the point of wrapping `rapidjson::FileReadStream` in another layer of abstraction here. It seems like the only thing we get for the trouble is a logging call when the file is closed, and it requires those who know rapidjson to read the source code to understand what's going on with the wrapper. Is there something that I'm not seeing?

tests/python/test_xgboost_integration.py
```diff
@@ -42,7 +42,9 @@ def test_xgb_boston(tmpdir, toolchain, objective, model_format):
         model_name = 'boston.json'
         model_path = os.path.join(tmpdir, model_name)
         bst.save_model(model_path)
-        model = treelite.Model.load(filename=model_path, model_format='xgboost_json')
+        with open(model_path, 'r') as f:
+            json_str = f.read()
+        model = treelite.Model.from_xgboost_json(json_str)
```
wphicks: Why not load directly from the file here? We have a later test that checks the equivalence of models deserialized in both ways, so why not use the more efficient one here?

src/frontend/xgboost_json.cc
```diff
@@ -23,16 +23,73 @@
...
+ public:
+  explicit OwningFileReadStream(const char* filename)
+      : fp_(OpenFile(filename)), file_stream_(nullptr), read_buffer_("") {
+    file_stream_ = std::make_unique<rapidjson::FileReadStream>(
+        fp_, read_buffer_, sizeof(read_buffer_));
+  }
+  virtual ~OwningFileReadStream() {
+    if (fp_) {
+      LOG(INFO) << "Closing file";
```
hcho3: This log should be removed.

src/frontend/xgboost_json.cc
hcho3: The point of the wrapper is to auto-close the file pointer. For the error handling, we need to close the file and open it again.

tests/python/test_xgboost_integration.py
hcho3: I changed it to test error handling. Let me revert this change.

src/frontend/xgboost_json.cc
hcho3: Actually I just found a way to remove the wrapper and still use error handling. See my latest commit.

src/frontend/xgboost_json.cc
```diff
@@ -42,27 +42,62 @@ namespace frontend {
...
-  return ParseStream(std::move(input_stream));
+  auto input_stream = std::make_unique<rapidjson::MemoryStream>(json_str, length);
+  auto error_handler = [json_str](size_t offset) -> std::string {
+    size_t cur = (offset >= 50 ? (offset - 50) : 0);
+    std::ostringstream oss, oss2;
+    for (int i = 0; i < 100; ++i) {
+      if (!json_str[cur]) {
+        break;
+      }
+      oss << json_str[cur];
```
wphicks, line 90: Any particular reason to break this up into two streams as opposed to just doing `oss << "\n"` and then continuing with everything you build into oss2?

wphicks: Wait, nevermind! Ignore this. I was looking at it wrong.

hcho3: `oss2` is for creating the second line `~~~~~~~~^~~~~~~~`, which is super useful to locate the exact location of the error.

wphicks (approved):
LGTM! Very helpful QOL improvement.

PR #232, opened 2020-12-05
Support XGBoost 1.3.0 JSON format

### LINKED ISSUE #226: XGBoost 1.3.0 support in the XGBoost JSON frontend
https://dev.azure.com/hcho3/treelite/_build/results?buildId=435&view=logs&j=e36176a6-f44b-587b-9f08-0a9ddc8a2920&t=912e1db4-e65f-567e-edbc-2b2aabd9d186

Treelite frontend currently crashes when given a model from XGBoost 1.3.0.

hcho3:
* Support XGBoost's brand-new categorical splits. XGBoost stores the list of categories that are associated with the **right** child node, unlike LightGBM which saves the list of categories associated with the left child node. Hence, we need to store extra boolean flag `categories_list_right_child_`.
* The JSON frontend should accept the extra fields `categories_segments`, `categories_sizes`, `categories_nodes`, and `categories`.

Closes #226


hcho3:
@trivialfis Is my observation correct? If a categorical value is found in the matching list, you'd traverse to the right child node?
https://github.com/dmlc/xgboost/blob/20c95be6252a7daee2facbcf7591dbb0e7e82283/src/common/categorical.h#L32-L43

wphicks (changes requested):
This all looks very good. It might be good to include tests for the new JSON features being parsed, but otherwise I think this all should be fine.

hcho3:
Good point. Let me create a test example from XGBoost.

PR #236, opened 2020-12-17
Fix C++ compiler warnings

hcho3:
This PR fixes C++ compiler warnings:
* Out-of-order initialization of struct members in the constructor
* Unused parameters
* Unused local variables
* Comparison between signed and unsigned variables

Also add a CMake option `ENABLE_ALL_WARNINGS` to enable all warnings from GCC/Clang.

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -347,7 +347,7 @@ class Tree {
    * \param nid ID of node being queried
    */
   inline std::vector<LeafOutputType> LeafVector(int nid) const {
-    if (nid > leaf_vector_offset_.Size()) {
+    if (static_cast<size_t>(nid) > leaf_vector_offset_.Size()) {
       throw std::runtime_error("nid too large");
     }
     return std::vector<LeafOutputType>(&leaf_vector_[leaf_vector_offset_[nid]],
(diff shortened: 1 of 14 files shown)


include/treelite/tree.h
```diff
@@ -347,7 +347,7 @@ class Tree {
    * \param nid ID of node being queried
    */
   inline std::vector<LeafOutputType> LeafVector(int nid) const {
-    if (nid > leaf_vector_offset_.Size()) {
+    if (static_cast<size_t>(nid) > leaf_vector_offset_.Size()) {
```
wphicks, line 350: I'll ask this question once, though I think it applies to most of the other places we're casting to `size_t`. Would it make more sense to have `nid` be a `size_t` to begin with? It's an array index which really should be a `size_t`, though I know that's a "purist" take that doesn't always play out in practice. Do we have a good reason not to make it a `size_t`?

wphicks (commented):
Just had a single question, but otherwise this looks great. Nice touch-up of a bunch of minor issues.

include/treelite/tree.h
hcho3: Let me try making it size_t.

hcho3: @wphicks It turns out it won't be wise to switch to `size_t`. Each tree node contains two integers to refer to the left and right child nodes, and switching to `size_t` will unnecessarily bloat the size of the model. Besides, we do not support trees with more than 2^31 nodes.

PR #251, opened 2021-02-02
Add support for "dart" booster in xgboost frontend

trevor-m:
https://xgboost.readthedocs.io/en/latest/tutorials/dart.html

Dart booster is very similar to gbtree at inference time, except that the leaf values of each tree are scaled by the "weight_drop" value. Currently, this implementation applies the scaling directly to the leaf value so that no changes are needed in the treelite backend.

Dart binary format is the same as GBTree but followed by weight_drop vector: https://github.com/dmlc/xgboost/blob/9f15b9e322ac505189394ff840bda3f52a7abb1f/src/gbm/gbtree.cc#L619-L622

Edit: Thanks wphicks for help with JSON frontend. I have implented that too now.
The dart json schema can be seen here: https://github.com/dmlc/xgboost/blob/master/src/gbm/gbtree.cc#L550-L561 
```
{
"learner": {
  "gradient_booster": {
    "name": "dart",
    "weight_drop": [... weight drop vector...],
    "gbtree": { "name": "gbtree", "model": ... the model which GBTreeModelHandler can parse }
  }
}
}
```

### DIFF
--- a/src/frontend/xgboost.cc
+++ b/src/frontend/xgboost.cc
@@ -367,9 +367,9 @@ inline std::unique_ptr<treelite::Model> ParseStream(dmlc::Stream* fi) {
   }
 
   /* loading GBTree */
-  CHECK_EQ(name_gbm_, "gbtree")
+  CHECK(name_gbm_ == "gbtree" || name_gbm_ == "dart")
     << "Invalid XGBoost model file: "
-    << "Gradient booster must be gbtree type.";
+    << "Gradient booster must be gbtree or dart type.";
 
   CHECK_EQ(fp->Read(&gbm_param_, sizeof(gbm_param_)), sizeof(gbm_param_))
     << "Invalid XGBoost model file: corrupted GBTree parameters";
@@ -378,6 +378,21 @@ inline std::unique_ptr<treelite::Model> ParseStream(dmlc::Stream* fi) {
     xgb_trees_.back().Load(fp.get());
   }
   CHECK_EQ(gbm_param_.num_roots, 1) << "multi-root trees not supported";
+  // tree_info is currently unused.
+  std::vector<int> tree_info;
+  tree_info.resize(gbm_param_.num_trees);
+  if (gbm_param_.num_trees != 0) {
+    CHECK_EQ(fp->Read(dmlc::BeginPtr(tree_info), sizeof(int32_t) * tree_info.size()),
+             sizeof(int32_t) * tree_info.size());
+  }
+  // Load weight drop values (per tree) for dart models.
+  std::vector<bst_float> weight_drop;
+  if (name_gbm_ == "dart") {
+    weight_drop.resize(gbm_param_.num_trees);
+    if (gbm_param_.num_trees != 0) {
+      fi->Read(&weight_drop);
+    }
+  }
 
   /* 2. Export model */
   std::unique_ptr<treelite::Model> model_ptr = treelite::Model::Create<float, float>();
@@ -427,7 +442,11 @@ inline std::unique_ptr<treelite::Model> ParseStream(dmlc::Stream* fi) {
       const XGBTree::Node& node = xgb_tree[old_id];
       const NodeStat stat = xgb_tree.Stat(old_id);
       if (node.is_leaf()) {
-        const bst_float leaf_value = node.leaf_value();
+        bst_float leaf_value = node.leaf_value();
+        // Fold weight drop into leaf value for dart models.
+        if (!weight_drop.empty()) {
+          leaf_value *= weight_drop[model->trees.size() - 1];
+        }
         tree.SetLeaf(new_id, static_cast<float>(leaf_value));
       } else {
         const bst_float split_cond = node.split_cond();
--- a/tests/python/test_xgboost_integration.py
+++ b/tests/python/test_xgboost_integration.py
@@ -292,9 +292,9 @@ def test_xgb_categorical_split(tmpdir, toolchain, quantize, parallel_comp):
     check_predictor(predictor, dataset)
 
 
-@pytest.mark.parametrize('model_format', ['binary'])
+@pytest.mark.parametrize('model_format', ['binary', 'json'])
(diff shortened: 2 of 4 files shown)


trevor-m:
@hcho3 Could you please review?

hcho3:
@trevor-m Hi, thank you for your contribution. I'm on vacation until March 3. In the meanwhile, please ask @wphicks for assistance with this pull request. He is the other committer for the Treelite project.

wphicks:
@trevor-m Yes! My apologies; I started to look at it and then got caught up in release stuff for another project. I'll give it a review soon.

trevor-m:
> @trevor-m Yes! My apologies; I started to look at it and then got caught up in release stuff for another project. I'll give it a review soon.

No worries! Please take your time.

tests/python/test_xgboost_integration.py
```diff
@@ -290,3 +290,53 @@ def test_xgb_categorical_split(tmpdir, toolchain, quantize, parallel_comp):
...
+             'skip_drop': 0.5}
+    bst = xgboost.train(param, dtrain=dtrain, num_boost_round=num_round)
+
+    if model_format == 'json':
+        model_name = 'dart.json'
+        model_path = os.path.join(tmpdir, model_name)
+        bst.save_model(model_path)
+        model = treelite.Model.load(filename=model_path, model_format='xgboost_json')
+    else:
+        model = treelite.Model.from_xgboost(bst)
```
wphicks: It would probably be good to follow the pattern established [here](https://github.com/dmlc/treelite/blob/mainline/tests/python/test_xgboost_integration.py#L206-L209) and save out the binary and reload it for precise parallelism with the json (de)serialization test. Alternatively, you could use `from_xgboost_json` in the json test, but I think it makes more sense to go through the full write/read cycle here.

wphicks (changes requested):
This looks great! One small tweak to the test, but otherwise everything looks perfect.

In terms of the JSON deserialization, the [implementation](https://github.com/dmlc/treelite/blob/mainline/src/frontend/xgboost_json.cc) is all in RapidJSON. I don't know if you've worked with that library before, but the basic pattern is that you have a handler that moves through each JSON element and delegates parsing to whatever more specialized handler is currently on top of the "stack" of delegator handlers. Typically when you start a new object, you push a specialized handler for that object onto the stack, and when you reach the end of that object, you pop the stack, returning control to the handler for the enclosing object.

So, you'll need to make 2 changes in the xgboost_json code to support dart. The simplest one is just adding `"dart"` as an allowed name [here](https://github.com/dmlc/treelite/blob/mainline/src/frontend/xgboost_json.cc#L337). The more important change will be adding a `GradientBoosterHandler::StartArray` method. When an array is encountered inside a `GradientBoosterHandler` object, you'll want to push a new handler onto the stack which will store the contents of that array to some temporary vector which can then be used to modify the leaf values as you've done with the binary case.

The most analogous example of this in the existing code is [here](https://github.com/dmlc/treelite/blob/mainline/src/frontend/xgboost_json.cc#L405-L408). In this case, all we need to do is store the version information and then check it [here](https://github.com/dmlc/treelite/blob/mainline/src/frontend/xgboost_json.cc#L433). In your case, you may need to think about where to store that vector a bit more in order to access it when the leaf values are constructed (or modify them after parsing is fully complete).

Let me know if that gives you enough to start on it or if you'd like some more detailed help.

trevor-m:
@wphicks Thank you for the detailed explanation, that was extremely helpful! I've implemented the json frontend support for dart now and updated the test case.

trevor-m:
@wphicks It looks like the java CI is failing. It looks like its for unrelated reasons, but please let me know if its related to my PR and I will try to fix it. https://dev.azure.com/hcho3/treelite/_build/results?buildId=512&view=logs&j=d249c4d1-3588-5000-63a8-8a1d89f50178&t=02a7030e-924a-5592-1359-e7cd0bcba7da&l=31407

wphicks:
Yeah, that looks pretty unrelated to me. I just kicked off a rerun, and we'll see what happens.

trevor-m:
@wphicks It appears related to Java version

https://github.com/dmlc/xgboost/issues/4000

https://discuss.xgboost.ai/t/problem-with-xgboost-installation-java-lang-noclassdeffounderror-javax-tools-toolprovider/437/2

wphicks:
@trevor-m Really sorry this has taken so long. I'm a relatively new reviewer on this project, and I've been trying to find someone on my team who has better familiarity with the Java CI infrastructure than I do to get this unstuck from CI. So far, no luck on that front, but I've reviewed everything else, and it all looks perfect. I'm going to mark this as approved, and we'll get it merged as soon as we get CI figured out, but there are definitely no further changes we'll need on your side.

wphicks (approved):

trevor-m:
@wphicks Thanks! No rush on my side.

src/frontend/xgboost.cc
```diff
@@ -378,6 +378,21 @@ inline std::unique_ptr<treelite::Model> ParseStream(dmlc::Stream* fi) {
     xgb_trees_.back().Load(fp.get());
   }
   CHECK_EQ(gbm_param_.num_roots, 1) << "multi-root trees not supported";
+  // tree_info is currently unused.
```
trivialfis, line 381: Out of curiosity, treelite doesn't have multi-class model?

hcho3:
I'm finally back from vacation. Will investigate the failing tests in Java.

src/frontend/xgboost.cc
hcho3: Do you mean multi-class classifiers? Treelite supports two methods for representing multi-class classifiers. See the docstring in https://github.com/dmlc/treelite/blob/49c35a52599d6728bf18552acddd5ee64fab419c/include/treelite/tree.h#L103-L121

Also see https://github.com/dmlc/treelite/issues/228

hcho3:
CI is fixed in #255

hcho3 (approved):

trevor-m:
Thanks @hcho3 !

PR #264, opened 2021-04-10
Faster model import for sklearn tree models

### LINKED ISSUE #263: Speed up loading time for sklearn RF models
The current method for loading scikit-learn random forest models is very slow even for moderately large forests. For example, the following code snippet takes 8 hours (!) to complete:
```python
from sklearn.datasets import make_regression
from sklearn.ensemble import RandomForestRegressor
from treelite.sklearn import import_model
import time

X, y = make_regression(n_samples=100000, n_features=100, n_informative=100)
clf = RandomForestRegressor(n_estimators=100, n_jobs=-1)
clf.fit(X, y)

tstart = time.perf_counter()
clf = import_model(clf)
tend = time.perf_counter()

print(f'Time elapsed = {tend - tstart} sec')
# 28723.539231207 sec on my machine
```

The `import_model` method should be made to run faster.

cc @wphicks @canonizer @JohnZed

hcho3:
Closes #263

Implement `treelite.sklearn.import_model_v2`, a faster method for importing sklearn models. The method is much faster than its predecessor (`treelite.sklearn.import_model`) because:
* The new method eschews the Model Builder API and builds a `treelite::Tree` object directly in C++.
* The new method avoids the use of recursive function calls when parsing the sklearn tree model. Instead, it takes in 7 flat arrays per tree.

Preliminary result: The following code snippet completes in **1.1 sec**. Since the same model would take 8 hours to load using the previous method, we have achieved 26122x speedup.
```python
import time
import ctypes
import numpy as np
from sklearn.datasets import make_regression
from sklearn.ensemble import RandomForestRegressor
from treelite.sklearn import import_model_v2

X, y = make_regression(n_samples=100000, n_features=100, n_informative=100)
clf = RandomForestRegressor(n_estimators=100, n_jobs=-1)
clf.fit(X, y)

tstart = time.perf_counter()
import_model_v2(clf)
tend = time.perf_counter()

print(f'Time elapsed = {tend - tstart} sec')
```

TODOs:
- [x] Implement `RandomForestRegressor`
- [x] Implement `RandomForestClassifier`
- [x] Implement `GradientBoostingRegressor`
- [x] Implement `GradientBoostingClassifier`
- [x] Implement `ExtraTreesRegressor`
- [x] Implement `ExtraTreesClassifier`
- [x] Add tests

cc @wphicks @canonizer @JohnZed


src/frontend/sklearn.cc
```diff
@@ -0,0 +1,235 @@
...
+    const int64_t total_sample_cnt = n_node_samples[tree_id][0];
+    while (!Q.empty()) {
+      int64_t node_id;
+      int new_node_id;
+      std::tie(node_id, new_node_id) = Q.front(); Q.pop();
+      const int64_t left_child_id = children_left[tree_id][node_id];
+      const int64_t right_child_id = children_right[tree_id][node_id];
+      const int64_t sample_cnt = n_node_samples[tree_id][node_id];
+      if (left_child_id == -1) {  // leaf node
+        // Get counts for each label (+/-) at this leaf node
```
hcho3: TODO: Refactor to eliminate code duplication.

include/treelite/c_api.h
```diff
@@ -171,6 +171,28 @@ TREELITE_DLL int TreeliteLoadXGBoostJSONString(const char* json_str,
 TREELITE_DLL int TreeliteLoadXGBoostModelFromMemoryBuffer(const void* buf,
                                                           size_t len,
                                                           ModelHandle* out);
+
+TREELITE_DLL int TreeliteLoadSKLearnRandomForestRegressor(
```
canonizer, line 200: Consider adding documentation that this method supports ExtraTrees as well.

src/frontend/sklearn.cc
```diff
@@ -0,0 +1,272 @@
...
+    const double** threshold, const double** value, const int64_t** n_node_samples,
+    const double** impurity, MetaHandlerFunc meta_handler, LeafHandlerFunc leaf_handler) {
+  CHECK_GT(n_trees, 0);
+  CHECK_GT(n_features, 0);
+
+  std::unique_ptr<treelite::Model> model_ptr = treelite::Model::Create<double, double>();
+  meta_handler(model_ptr.get(), n_features, n_classes);
+  auto* model = dynamic_cast<treelite::ModelImpl<double, double>*>(model_ptr.get());
+
+  for (int tree_id = 0; tree_id < n_trees; ++tree_id) {
```
canonizer, line 31: Optional: consider OpenMP parallelization. This is also fine to do in a separate pull request.

python/treelite/sklearn/__init__.py
```diff
@@ -97,4 +106,131 @@ class SKLRFMultiClassifierConverter(SKLRFMultiClassifierMixin, SKLConverterBase)
...
+                    f'Expected shape: {expected_shape}, Got shape {array.shape}'
+        v = np.array(array, copy=False, dtype=self.dtype, order='C')
+        self.collection.append(v.ctypes.data_as(self.ptr_type))
+
+    def as_c_array(self):
+        """Prepare the collection to pass as an argument of a C function"""
+        return c_array(self.ptr_type, self.collection)
+
+
+def import_model_v2(sklearn_model):
```
canonizer: Why not use the `import_model` name for the new function, and rename the old one to something like `import_model_old`?

canonizer (approved):
Approved.

Possibly for a different pull request: would it be possible to import other Scikit-Learn tree-based models from https://scikit-learn.org/stable/modules/classes.html#module-sklearn.ensemble as well?

python/treelite/sklearn/__init__.py
```diff
@@ -1,7 +1,14 @@
 # coding: utf-8
 """Converter to ingest scikit-learn models into Treelite"""
 
+from __future__ import absolute_import as _abs
```
wphicks: Why do we need this here?

python/treelite/sklearn/__init__.py
```diff
@@ -97,4 +106,131 @@ class SKLRFMultiClassifierConverter(SKLRFMultiClassifierMixin, SKLConverterBase)
...
+        Python handle to scikit-learn model
+
+    Returns
+    -------
+    model : :py:class:`~treelite.Model` object
+        loaded model
+    """
+    class_name = sklearn_model.__class__.__name__
+    module_name = sklearn_model.__module__.split('.')[0]
+    if module_name != 'sklearn':
```
wphicks: Is there some reason we're checking the type like this as opposed to doing a direct `isinstance` call with `BaseEstimator` or similar?

python/treelite/sklearn/__init__.py
```diff
@@ -97,4 +106,131 @@ class SKLRFMultiClassifierConverter(SKLRFMultiClassifierMixin, SKLConverterBase)
...
+        raise TreeliteError('Not a scikit-learn model')
+
+    if class_name in ['RandomForestRegressor', 'ExtraTreesRegressor', 'GradientBoostingRegressor',
```
wphicks: Same general question here. Any reason not to compare directly to the `sklearn` classes?

wphicks (commented):
In general, this looks great! My only significant questions here are around how we're doing type inspection on the Python side.

python/treelite/sklearn/__init__.py
hcho3: Ah, there are several occurrences of `__future__` that are now not needed, since we dropped support for Python 2 a while ago.

python/treelite/sklearn/__init__.py
hcho3: This was done to keep `sklearn` optional. Would it be better if we directly `assert` the existing of `sklearn` module?

src/frontend/sklearn.cc
hcho3: Sure. Will do in a future pull request.

python/treelite/sklearn/__init__.py
hcho3: Points taken. Will rename the function.

include/treelite/c_api.h
hcho3: Got it. I will add docstrings to all functions introduced in this pull request.

python/treelite/sklearn/__init__.py
wphicks: If an sklearn model is passed into this function, then `sklearn` should be available right? I think I'd favor a guarded sklearn import and `isinstance` checks to make this a bit more robust. I don't think there's any need to perform an assertion. We can just try to use it and handle the exception appropriately.

PR #270, opened 2021-04-28
Prototype for binary serializer to a file stream

### LINKED ISSUE #265: (De)serialization method of Treelite model object
To make it convenient for downstream projects (e.g. cuML) that use the Treelite model representation to consume decision tree models, we should add a method to serialize and deserialize the Treelite model object.

This way, downstream projects may avoid depending on the Python runtime, if it were to load tree models from cuML or scikit-learn.

hcho3:
Closes #265 

TODOs
- [x] Expose a C API to (de)serialize to a file stream
- [x] Write a small utility script that lets users convert a scikit-learn pickle into a Treelite binary serialized file
- [x] Write an end-to-end test


hcho3:
Still need to write a CLI wrapper script for the serializer, but functionality-wise this is ready for review.

include/treelite/tree_impl.h
```diff
@@ -46,6 +46,69 @@ inline std::string GetString<double>(double x) {
 
 namespace treelite {
 
+inline void
+PyBufferFrame::Serialize(FILE* dest_fp) const {
+  auto write_to_file = [](const void* buffer, size_t size, size_t count, FILE* fp) {
```
wphicks: It would be preferable to use `std::size_t` here and include `<cstddef>`, but if that is not possible, we at least need to include `<stddef.h>`

include/treelite/tree.h
```diff
@@ -35,13 +36,26 @@ float stof(const std::string& value, size_t* pos);
...
+// Represent a frame in the Python buffer protocol (PEP 3118). We use a simplified representation
+// to hold only 1-D arrays with stride 1.
 struct PyBufferFrame {
   void* buf;
   char* format;
   size_t itemsize;
   size_t nitem;
+
+  // Serialize a frame to a file stream
+  inline void Serialize(FILE* dest_fp) const;
```
wphicks: Can you remind me why we need to mark this as inline? I remember thinking about this before and concluding that ODR required it for functions in here, but I'm not immediately seeing the ODR justification when I'm looking at it this time.

include/treelite/tree_impl.h
```diff
@@ -418,10 +492,14 @@ inline void InitScalarFromPyBuffer(TaskType* scalar, PyBufferFrame buffer) {
   }
   T* t = static_cast<T*>(buffer.buf);
   *scalar = static_cast<TaskType>(*t);
+  if (assume_ownership) {
```
wphicks: We're doing quite a bit of manual ownership tracking throughout. Would it be possible to use built-in ownership semantics? I.e. If we want to transfer ownership, move a `unique_ptr`, if we want to give access without sharing, provide a `weak_ptr`, etc.?

include/treelite/tree_impl.h
```diff
@@ -705,26 +788,71 @@ Model::Dispatch(Func func) const {
...
+  InitScalarFromPyBuffer(&leaf_output_type, frames[4], assume_ownership);
 
   std::unique_ptr<Model> model = Model::Create(threshold_type, leaf_output_type);
-  model->InitFromPyBuffer(frames.begin() + 2, frames.end());
+  model->InitFromPyBuffer(frames.begin() + kNumFrameInHeader, frames.end(), assume_ownership);
   return model;
 }
 
+inline std::unique_ptr<Model>
+Model::CreateFromPyBuffer(std::vector<PyBufferFrame> frames) {
```
wphicks: Is the copy necessary here? Is there a scenario in which someone would call `CreateFromPyBuffer(frames)` and then continue to use those frames later? Could we just accept `std::vector<PyBufferFrame>&&` instead?

wphicks (commented):
This looks really great! The only thing I'm squeamish about is all of the manual memory management we're doing throughout. If possible, I'd love to see us use a little more RAII and indicate ownership through owning/non-owning pointers where needed, but maybe there's some reason for not doing that that I'm not aware of in this case?

include/treelite/tree.h
hcho3: The body of this function is defined in `tree_impl.h`.

hcho3:
> The only thing I'm squeamish about is all of the manual memory management we're doing throughout. If possible, I'd love to see us use a little more RAII and indicate ownership through owning/non-owning pointers where needed

The limitation is `PyBufferFrame` struct, which contains two raw pointers `buf` and `format`. These pointers refer to buffers that might originate from other libraries, e.g. the Python interpreter. The Python buffer protocol ([PEP 3118](https://www.python.org/dev/peps/pep-3118/)) was designed to allow multiple libraries to exchange raw buffers.

We could avoid using `PyBufferFrame` entirely for this PR, but then we'd need to duplicate all the serialization logic that's available in `GetPyBuffer()` and `CreateFromPyBuffer()`. The logic was originally written to support in-memory serialization, but it's also useful for serializing to the disk.

hcho3:
I added a CLI script.

Usage:
```
python -m treelite.serialize --input-model ./model.pkl --input-model-type sklearn_pkl  \
    --output-checkpoint saved.tl
```

Help:
```
usage: __main__.py [-h] --input-model INPUT_MODEL --input-model-type {sklearn_pkl,xgboost,xgboost_json,lightgbm}
                   --output-checkpoint OUTPUT_CHECKPOINT

CLI script to convert a model file to a Treelite binary checkpoint

optional arguments:
  -h, --help            show this help message and exit
  --input-model INPUT_MODEL
                        Path to the tree model file
  --input-model-type {sklearn_pkl,xgboost,xgboost_json,lightgbm}
                        Type of the tree model file
  --output-checkpoint OUTPUT_CHECKPOINT
                        Path to the checkpoint file, using a special binary format for Treelite
```

include/treelite/tree_impl.h
hcho3: Done

include/treelite/tree_impl.h
hcho3: > Is there a scenario in which someone would call `CreateFromPyBuffer(frames)` and then continue to use those frames later?

Yes, the `frames` can originate from outside. For example, in MNMG setting, each worker process receives a series of frames that represent a cuML RF model. The frames are held by the Python interpreter, and our C++ code has no control over when they will be used or destroyed.

include/treelite/tree.h
hcho3: I moved `Serialize()` to its own source file, so it no longer has the `inline` designation.

include/treelite/tree_impl.h
hcho3: I redesigned the serialization logic so that the file serializer doesn't depend at all on `PyBufferFrame`, so we don't need to do manual ownership tracking.

hcho3:
I redesigned the serialization code so that the file serializer doesn't depend at all on `PyBufferFrame`, so manual ownership tracking is no longer required. I avoided code duplication by factoring out common serialization logic as template functions.

@wphicks Can I get another round of review?

wphicks (approved):
Looks great! Can't wait to make use of it

PR #273, opened 2021-05-06
Hotfix for serializer

hcho3:
Failing tests in rapidsai/cuml#3829 revealed a subtle bug in #270. It appears that the testing suite in Treelite was not sufficiently robust to let this bug through 😞

The bug causes multiple trees to deserialize from the same set of frames.

I added a few test cases that are affected by the bug. The current Treelite version will fail these tests.

Once this PR is merged, I will publish a patch release (1.2.1).

### DIFF
--- a/include/treelite/tree_impl.h
+++ b/include/treelite/tree_impl.h
@@ -925,6 +925,7 @@ ModelImpl<ThresholdType, LeafOutputType>::InitFromPyBuffer(
 
   auto tree_hanlder = [&begin](Tree<ThresholdType, LeafOutputType>& tree) {
     tree.InitFromPyBuffer(begin, begin + kNumFramePerTree);
+    begin += kNumFramePerTree;
   };
 
   DeserializeTemplate(num_tree, header_field_handler, tree_hanlder);
--- a/include/treelite/tree_impl.h
+++ b/include/treelite/tree_impl.h
@@ -859,7 +859,7 @@ template <typename ThresholdType, typename LeafOutputType>
 template <typename HeaderFieldHandlerFunc, typename TreeHandlerFunc>
 inline void
 ModelImpl<ThresholdType, LeafOutputType>::DeserializeTemplate(
-    size_t num_tree,
+    std::size_t num_tree,
     HeaderFieldHandlerFunc header_field_handler,
     TreeHandlerFunc tree_handler) {
   /* Header */
@@ -870,7 +870,7 @@ ModelImpl<ThresholdType, LeafOutputType>::DeserializeTemplate(
   header_field_handler(&param);
   /* Body */
   trees.clear();
-  for (size_t i = 0; i < num_tree; ++i) {
+  for (std::size_t i = 0; i < num_tree; ++i) {
     trees.emplace_back();
     tree_handler(trees.back());
   }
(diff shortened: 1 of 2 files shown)


tests/cpp/test_serializer.cc
```diff
@@ -44,7 +44,7 @@ inline void TestRoundTrip(treelite::Model* model) {
     std::unique_ptr<treelite::Model> received_model = treelite::Model::DeserializeFromFile(fp);
     std::fclose(fp);
 
-    ASSERT_EQ(TreeliteToBytes(model), TreeliteToBytes(received_model.get()));
+    ASSERT_TRUE(TreeliteToBytes(model) == TreeliteToBytes(received_model.get()));
```
hcho3, line 51: `ASSERT_EQ` will dump all the raw bytes into a string, leading to OOM error for this test case. So use `ASSERT_TRUE` instead.

tests/cpp/test_serializer.cc
```diff
@@ -189,10 +189,10 @@ void PyBufferInterfaceRoundTrip_TreeDepth2() {
...
     tree->SetCategoricalTestNode(2, 1, {0}, true, 5, 6);
     tree->SetRootNode(0);
-    tree->SetLeafNode(3, frontend::Value::Create<LeafOutputType>(3));
-    tree->SetLeafNode(4, frontend::Value::Create<LeafOutputType>(1));
-    tree->SetLeafNode(5, frontend::Value::Create<LeafOutputType>(4));
-    tree->SetLeafNode(6, frontend::Value::Create<LeafOutputType>(2));
+    tree->SetLeafNode(3, frontend::Value::Create<LeafOutputType>(tree_id + 3));
+    tree->SetLeafNode(4, frontend::Value::Create<LeafOutputType>(tree_id + 1));
+    tree->SetLeafNode(5, frontend::Value::Create<LeafOutputType>(tree_id + 4));
+    tree->SetLeafNode(6, frontend::Value::Create<LeafOutputType>(tree_id + 2));
```
hcho3, line 199: I crafted the test case so that all the trees will have different leaf outputs. The bug being addressed causes all trees to be identical.

include/treelite/tree_impl.h
```diff
@@ -925,6 +925,7 @@ ModelImpl<ThresholdType, LeafOutputType>::InitFromPyBuffer(
 
   auto tree_hanlder = [&begin](Tree<ThresholdType, LeafOutputType>& tree) {
     tree.InitFromPyBuffer(begin, begin + kNumFramePerTree);
+    begin += kNumFramePerTree;
```
wphicks, line 929: Since we're capturing `begin` by reference, won't this iterate begin forward as a side effect of executing the lambda? It's not immediately clear to me why that is the correct thing to do with this handler.  Maybe add a comment or refactor such that the lambda returns the next iterator to be processed rather than directly changing `begin`?

tests/cpp/test_serializer.cc
wphicks: Maybe add a comment to this effect so that someone doesn't naively reintroduce this problem down the line?

wphicks (changes requested):

include/treelite/tree_impl.h
hcho3: > Since we're capturing begin by reference, won't this iterate begin forward as a side effect of executing the lambda?

Yes, this is intended behavior.

> refactor such that the lambda returns the next iterator to be processed

That would not be feasible, since the tree handler for the file stream does not involve iterators.

wphicks: Great! Can we get a comment just to make it clear to future developers that the side-effect is intentional?

wphicks (approved):

PR #274, opened 2021-05-12
[EXPERIMENTAL] Add GTIL, reference inference backend

### LINKED ISSUE #11: load model from file and predict
I'm interested in using `treelite` as a fast and lightweight C/C++ prediction module.  It seems the preferred model is to load a standard supported format, compile/bake an optimized c file/library, and then integrate the generated c file into an application for prediction.  Is it possible to use `treelite` to load a model from file at runtime for prediction?  (If this isn't possible through the current API would you accept this feature).  This would provide some flexibility and code size reductions compared to using `xgboost` or `lightgbm` directly.

hcho3:
General Tree Inference Library (GTIL) provides a reference implementation for predicting with decision trees. GTIL is useful in cases it is infeasible to build the tree models as native shared libs.

Fixes https://github.com/dmlc/treelite/issues/11


wphicks (approved):
Just finished going over this, and I have no suggestions; it looks great!

PR #286, opened 2021-07-08
[CI] Reduce env setup time in Travis CI

hcho3:
Avoid running `brew update` to save the CI time on Travis CI.


hcho3:
This PR didn't actually reduce the setup time. Will think of alternatives.

PR #285, opened 2021-07-08
[Breaking] Remove dmlc-core dependency

hcho3:
Closes #244

This PR will be divided into smaller PRs for easier review.

Marking this PR as "breaking," per discussion in #284

PR #287, opened 2021-07-09
Remove dmlc logger

hcho3:
Extracted from #285. Re-implement logging functionalities such as `CHECK` and `LOG`, so that we no longer depend on the header `dmlc/logging.h`.

### DIFF
--- a/include/treelite/logging.h
+++ b/include/treelite/logging.h
@@ -9,8 +9,143 @@
 
 #include <treelite/thread_local.h>
 #include <iostream>
+#include <stdexcept>
+#include <string>
+#include <sstream>
+#include <memory>
+#include <cstdio>
+#include <ctime>
 
 namespace treelite {
+
+/*!
+ * \brief Exception class that will be thrown by Treelite
+ */
+struct Error : public std::runtime_error {
+  explicit Error(const std::string& s) : std::runtime_error(s) {}
+};
+
+template <typename X, typename Y>
+std::unique_ptr<std::string> LogCheckFormat(const X& x, const Y& y) {
+  std::ostringstream os;
+  os << " (" << x << " vs. " << y << ") ";
+  /* CHECK_XX(x, y) requires x and y can be serialized to string. Use CHECK(x OP y) otherwise. */
+  return std::make_unique<std::string>(os.str());
+}
+
+#if defined(__GNUC__) || defined(__clang__)
+#define TREELITE_ALWAYS_INLINE inline __attribute__((__always_inline__))
+#elif defined(_MSC_VER)
+#define TREELITE_ALWAYS_INLINE __forceinline
+#else
+#define TREELITE_ALWAYS_INLINE inline
+#endif
+
+#define DEFINE_CHECK_FUNC(name, op)                                                        \
+  template <typename X, typename Y>                                                        \
+  TREELITE_ALWAYS_INLINE std::unique_ptr<std::string> LogCheck##name(const X& x, const Y& y) { \
+    if (x op y) return nullptr;                                                            \
+    return LogCheckFormat(x, y);                                                           \
+  }                                                                                        \
+  TREELITE_ALWAYS_INLINE std::unique_ptr<std::string> LogCheck##name(int x, int y) {           \
+    return LogCheck##name<int, int>(x, y);                                                 \
+  }
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsign-compare"
+DEFINE_CHECK_FUNC(_LT, <)
+DEFINE_CHECK_FUNC(_GT, >)
+DEFINE_CHECK_FUNC(_LE, <=)
+DEFINE_CHECK_FUNC(_GE, >=)
+DEFINE_CHECK_FUNC(_EQ, ==)
+DEFINE_CHECK_FUNC(_NE, !=)
+#pragma GCC diagnostic pop
+
+
+#define CHECK_BINARY_OP(name, op, x, y)                  \
+  if (auto __treelite__log__err = treelite::LogCheck##name(x, y))  \
+      treelite::LogMessageFatal(__FILE__, __LINE__).stream() \
+        << "Check failed: " << #x " " #op " " #y << *__treelite__log__err << ": "
+#define CHECK(x)                                               \
+  if (!(x))                                                    \
+    treelite::LogMessageFatal(__FILE__, __LINE__).stream()     \
+      << "Check failed: " #x << ": "
+#define CHECK_LT(x, y) CHECK_BINARY_OP(_LT, <, x, y)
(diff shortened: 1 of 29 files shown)


include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+#else  // defined(_MSC_VER)
+    time_t time_value = std::time(nullptr);
+    struct tm* pnow;
+#if !defined(_WIN32)
+    struct tm now;
+    pnow = localtime_r(&time_value, &now);
+#else  // !defined(_WIN32)
+    pnow = std::localtime(&time_value);  // NOLINT(*)
+#endif  // !defined(_WIN32)
+    std::snprintf(buffer_, sizeof(buffer_), "%02d:%02d:%02d",
```
trivialfis, line 104: Is there a type in c++ chrono for these?

include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+
+#define CHECK_BINARY_OP(name, op, x, y)                  \
+  if (auto __treelite__log__err = treelite::LogCheck##name(x, y))  \
+      treelite::LogMessageFatal(__FILE__, __LINE__).stream() \
+        << "Check failed: " << #x " " #op " " #y << *__treelite__log__err << ": "
+#define CHECK(x)                                               \
+  if (!(x))                                                    \
+    treelite::LogMessageFatal(__FILE__, __LINE__).stream()     \
+      << "Check failed: " #x << ": "
+#define CHECK_LT(x, y) CHECK_BINARY_OP(_LT, <, x, y)
```
trivialfis, line 73: I'm not sure if this is used in runtime, there was issue around xgboost that users want to use the c++ header and other libraries also use similar macros.

include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+DEFINE_CHECK_FUNC(_LE, <=)
+DEFINE_CHECK_FUNC(_GE, >=)
+DEFINE_CHECK_FUNC(_EQ, ==)
+DEFINE_CHECK_FUNC(_NE, !=)
+#pragma GCC diagnostic pop
+
+
+#define CHECK_BINARY_OP(name, op, x, y)                  \
+  if (auto __treelite__log__err = treelite::LogCheck##name(x, y))  \
+      treelite::LogMessageFatal(__FILE__, __LINE__).stream() \
```
trivialfis: Might a a good idea to use `::treelite:: ...` to specify we should start with global namespace.

include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+    treelite::LogMessageFatal(__FILE__, __LINE__).stream()     \
+      << "Check failed: " #x << ": "
+#define CHECK_LT(x, y) CHECK_BINARY_OP(_LT, <, x, y)
+#define CHECK_GT(x, y) CHECK_BINARY_OP(_GT, >, x, y)
+#define CHECK_LE(x, y) CHECK_BINARY_OP(_LE, <=, x, y)
+#define CHECK_GE(x, y) CHECK_BINARY_OP(_GE, >=, x, y)
+#define CHECK_EQ(x, y) CHECK_BINARY_OP(_EQ, ==, x, y)
+#define CHECK_NE(x, y) CHECK_BINARY_OP(_NE, !=, x, y)
+
+#define LOG_INFO treelite::LogMessage(__FILE__, __LINE__)
```
trivialfis: For the future, maybe a good idea to distinguish user error and internal error.

include/treelite/logging.h
hcho3: Should we rename all uses of LOG to something like TREELITE_LOG to avoid collision? RapidJSON uses RAPIDJSON_ASSERT.

trivialfis: Depends on whether user might include this file.  If it's not in the runtime then it should be fine.  Besides we can think about it in the future to keep this PR in scope. ;-)

hcho3: Got it, I'll do the renaming in a future PR.

include/treelite/logging.h
hcho3: No, chrono doesn't deal with date: https://stackoverflow.com/questions/17223096/outputting-date-and-time-in-c-using-stdchrono. We need to use std::localtime or std::localtime_r.

trivialfis: That's unfortunate.

hcho3: C++20 provides additional date utilities, but we can't use C++20 yet.

include/treelite/logging.h
hcho3: Filed #294

include/treelite/logging.h
hcho3: Done

include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+#pragma GCC diagnostic pop
+
+
+#define CHECK_BINARY_OP(name, op, x, y)                  \
+  if (auto __treelite__log__err = treelite::LogCheck##name(x, y))  \
+      ::treelite::LogMessageFatal(__FILE__, __LINE__).stream() \
+        << "Check failed: " << #x " " #op " " #y << *__treelite__log__err << ": "
+#define CHECK(x)                                               \
+  if (!(x))                                                    \
+    treelite::LogMessageFatal(__FILE__, __LINE__).stream()     \
```
trivialfis: `::treelite::`.

include/treelite/logging.h
trivialfis: `::treelite::`.

include/treelite/logging.h
```diff
@@ -9,9 +9,144 @@
...
+#define CHECK_LT(x, y) CHECK_BINARY_OP(_LT, <, x, y)
+#define CHECK_GT(x, y) CHECK_BINARY_OP(_GT, >, x, y)
+#define CHECK_LE(x, y) CHECK_BINARY_OP(_LE, <=, x, y)
+#define CHECK_GE(x, y) CHECK_BINARY_OP(_GE, >=, x, y)
+#define CHECK_EQ(x, y) CHECK_BINARY_OP(_EQ, ==, x, y)
+#define CHECK_NE(x, y) CHECK_BINARY_OP(_NE, !=, x, y)
+
+#define LOG_INFO treelite::LogMessage(__FILE__, __LINE__)
+#define LOG_ERROR LOG_INFO
+#define LOG_FATAL treelite::LogMessageFatal(__FILE__, __LINE__)
```
trivialfis: `::treelite::`.

trivialfis (approved):

hcho3:
@trivialfis Thank you so much for reviewing the 7 pull requests! I was worried about the size of the changes required to remove `dmlc`, but thanks to your review, I can merge it with a peace of mind.

PR #288, opened 2021-07-09
Remove the use of dmlc object registry

hcho3:
Extracted from #285.

* Do not use the header `dmlc/registry.h`.
* Do not use the object registry provided by dmlc-core. The registry lets you instantiate a new object by the class name. We really don't need the registry, since there are only two kinds of code generators (`FailSafeCompiler` and `ASTNativeCompiler`). Simple `if` / `else` statements suffice.
* Use pimpl pattern to hide templated functions in the code generators.
* Define new headers `failsafe.h` and `ast_native.h` to store the class declarations of `FailSafeCompiler` and `ASTNativeCompiler`. This is more in line with conventional C++.
* Rename `param` to `param_`.


src/compiler/failsafe.h
```diff
@@ -0,0 +1,32 @@
...
+
+class FailSafeCompilerImpl;
+
+class FailSafeCompiler : public Compiler {
+ public:
+  explicit FailSafeCompiler(const CompilerParam& param);
+  CompiledModel Compile(const Model& model) override;
+ private:
+  std::unique_ptr<FailSafeCompilerImpl> pimpl_;
+};
```
trivialfis, line 29: Interesting, I always find using unique ptr with incomplete class troublesome due to missing destructor.  Not sure how did you get it to work.

src/compiler/failsafe.cc
```diff
@@ -389,17 +367,40 @@ class FailSafeCompiler : public Compiler {
...
 
-TREELITE_REGISTER_COMPILER(FailSafeCompiler, "failsafe")
-.describe("Simple compiler to express trees as a tight for-loop")
-.set_body([](const CompilerParam& param) -> Compiler* {
-    return new FailSafeCompiler(param);
-  });
+FailSafeCompiler::FailSafeCompiler(const CompilerParam& param)
+    : pimpl_(std::make_unique<FailSafeCompilerImpl>(param)) {
+  if (param.verbose > 0) {
+    LOG(INFO) << "Using FailSafeCompiler";
```
trivialfis, line 400: Is there a log warning macro?

hcho3: LOG(WARN) is just an alias of LOG(INFO). Treelite doesn't let you set global verbosity level like XGBoost.

src/compiler/failsafe.h
hcho3: Take a look at https://www.fluentcpp.com/2017/09/22/make-pimpl-using-unique_ptr/. The key is to explicitly define the destructor for `FailSafeCompiler` after the full definition of `FailSafeCompilerImpl` is provided. See my latest commit.

src/compiler/failsafe.cc
trivialfis: Got it.

src/compiler/failsafe.h
trivialfis: Thanks for the reference.  That's a weird aspect of c++.

trivialfis (approved):

PR #289, opened 2021-07-09
Remove the use of dmlc::JSONWriter, dmlc::Stream, and dmlc serializer

hcho3:
Extracted from #285.

* Do not use `dmlc/json.h`, `dmlc/io.h`, and `dmlc/serializer.h`.
* Replace all uses of `dmlc::Stream` with `std::fstream` and `std::stringstream`.
* Replace `dmlc::JSONWriter` with `rapidjson::Writer`.
* Use `uint64_t` for storing instance counts in JSON.
* Rename `TaskParameter` to `TaskParam`, to be consistent with `ModelParam`.
* Replace `ReferenceSerialize` with `SerializeToJSON`. Previously, `ReferenceSerialize` was a reference implementation of serializer for tree models. `ReferenceSerialize` was useful for verifying the correctness of other serializer methods of the tree objects. The reference serializer had to go, unfortunately, because it used dmlc serializer functions. As a replacement, I implemented a JSON serializer `SerializeToJSON`. While not very efficient compared to binary serializers, the JSON serializer will be useful for debugging and testing.
* Reimplement `LoadText` with `std::istream` and `std::getline`.


include/treelite/annotator.h
```diff
@@ -27,14 +29,14 @@ class BranchAnnotator {
   void Annotate(const Model& model, const DMatrix* dmat, int nthread, int verbose);
   /*!
    * \brief load branch annotation from a JSON file
-   * \param fi input stream
+   * \param fp input stream
    */
-  void Load(dmlc::Stream* fi);
+  void Load(FILE* fp);
```
trivialfis: Why not c++ stream?

src/frontend/xgboost.cc
```diff
@@ -44,12 +43,21 @@ namespace {
 
 typedef float bst_float;
 
+template <typename T>
+inline T* BeginPtr(std::vector<T>& vec) {  // NOLINT(*)
+  if (vec.empty()) {
```
trivialfis: Can you use the data member function?

trivialfis (commented):
Looked briefly.  Not a complete review.

include/treelite/annotator.h
hcho3: Good point. Let me use `std::istream&` here, to be consistent with the changes made in `frontend/xgboost.cc` and `frontend/lightgbm.cc`.

src/frontend/xgboost.cc
hcho3: Done. All uses of `BeginPtr` in `frontend/xgboost.cc` involve non-empty containers, so we can just use `data()` instead.

trivialfis (approved):

PR #291, opened 2021-07-09
Remove use of dmlc::thread_local

hcho3:
Extracted from #285.

* Do not use `dmlc/thread_local.h`.
* Replace `dmlc::ThreadLocalStore` with `treelite::ThreadLocalStore`, which is a light wrapper around `thread_local` storage.


trivialfis (approved):
LGTM

PR #293, opened 2021-07-09
[Breaking] Remove dmlc::Parameter

hcho3:
Extracted from #285.

* Do not use `dmlc/parameter.h`.
* `CompilerParam` no longer inherits from `dmlc::Parameter`. Consequently, it now implements a special method `ParseFromJSON()` that ingests parameters from a JSON string.
* Now parameters are set with JSON strings.
Before:
```cpp
CompilerHandle handle;
TreeliteCompilerCreate("ast_native", &handle);
TreeliteCompilerSetParam(handle, "quantize", "1");
TreeliteCompilerSetParam(handle, "parallel_comp", "100");
TreeliteCompilerSetParam(handle, "native_lib_name", "predictor");
```
After:
```cpp
CompilerHandle handle;
const char* params = "{\"quantize\": 1, \"parallel_comp\": 100,"
                     " \"native_lib_name\": \"predictor\"}";
TreeliteCompilerCreateV2("ast_native", params, &handle);
```

* Breaking: The C API functions `TreeliteCompilerCreate` and `TreeliteCompilerGenerateCode` are renamed to `TreeliteCompilerCreateV2` and `TreeliteCompilerGenerateCodeV2`, respectively, to indicate the different calling signature. In addition, `TreeliteCompilerSetParam` has been removed.
* Add C++ tests to test the correctness of `ParseFromJSON()`.
* Upgrade Google Test framework to 1.11.0 to test the content of exceptions.

### DIFF
--- a/python/treelite/frontend.py
+++ b/python/treelite/frontend.py
@@ -5,6 +5,7 @@ import ctypes
 import collections
 import shutil
 import os
+import json
 from tempfile import TemporaryDirectory
 
 import numpy as np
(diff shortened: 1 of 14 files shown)


python/treelite/frontend.py
```diff
@@ -304,40 +305,19 @@ def compile(self, dirpath, params=None, compiler='ast_native', verbose=False):
...
         """
         compiler_handle = ctypes.c_void_p()
-        _check_call(_LIB.TreeliteCompilerCreate(c_str(compiler),
-                                                ctypes.byref(compiler_handle)))
         _params = dict(params) if isinstance(params, list) else params
-        self._set_compiler_param(compiler_handle, _params or {})
-        _check_call(_LIB.TreeliteCompilerGenerateCode(
+        if verbose and _params:
+            _params['verbose'] = 1
+        params_json_str = json.dumps(_params) if _params else '{}'
```
trivialfis, line 311: I think JSON handles this automatically.

hcho3: No, the None object gets serialized to `null`.

trivialfis: Got it, the param is optional.

trivialfis (approved):

PR #299, opened 2021-07-15
[CI] Downgrade libomp package to 11.1.0

hcho3:
https://app.travis-ci.com/github/dmlc/treelite/jobs/524580900#L3132:
```
tests/python/test_basic.py::test_basic[mushroom-True-None-True-gcc] 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong
with the build itself.
Check the details on how to adjust your build configuration on:
https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

I suspect it has to do with the use of OpenMP in `src/annotator.cc`.

PR #300, opened 2021-07-16
Remove OpenMP dependency

hcho3:
Follow-up to #299.

On MacOS, the latest OpenMP runtime (`libomp`) is causing Treelite to hang. Fortunately, it is relatively painless to remove the use of OpenMP from Treelite, as only one feature of Treelite (branch annotator) uses it. (The Treelite runtime uses a custom thread pool.)

Replace OpenMP with C++11 threads.

As a side effect, this PR also speeds up testing on MacOS, because we no longer have to install packages from Homebrew. This shaves off 15 min of CI time.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -11,17 +11,6 @@ endif(UNIX)
 
 add_library(objtreelite_common OBJECT)  # Component shared by both main package and runtime
 
-if(USE_OPENMP)
-  if (APPLE)
-    # Require CMake 3.16+ on Mac OSX, as previous versions of CMake had trouble locating
-    # OpenMP on Mac. See https://github.com/dmlc/xgboost/pull/5146#issuecomment-568312706
-    cmake_minimum_required(VERSION 3.16)
-  endif (APPLE)
-  find_package(OpenMP REQUIRED)
-else()
-  message(STATUS "Disabling OpenMP")
-endif()
-
 if(ENABLE_ALL_WARNINGS)
   foreach(target objtreelite objtreelite_runtime objtreelite_runtime)
     target_compile_options(${target} PRIVATE -Wall -Wextra)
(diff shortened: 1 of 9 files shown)


src/CMakeLists.txt
```diff
@@ -33,17 +24,14 @@ foreach(lib objtreelite objtreelite_runtime objtreelite_common)
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
     $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
     $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
+  target_link_libraries(${lib} PUBLIC Threads::Threads)
```
trivialfis, line 27: Do we need to link `${CMAKE_THREAD_LIBS_INIT}`?

src/threading_utils/parallel_for.h
```diff
@@ -0,0 +1,83 @@
...
+
+template <typename IndexType, typename FuncType>
+void ParallelFor(IndexType begin, IndexType end, std::size_t nthread, FuncType func) {
+  TREELITE_CHECK_GT(nthread, 0) << "nthread must be positive";
+  TREELITE_CHECK_LE(nthread, std::thread::hardware_concurrency())
+    << "nthread cannot exceed " << std::thread::hardware_concurrency();
+  if (begin == end) {
+    return;
+  }
+  /* Divide the rnage [begin, end) equally among the threads.
```
trivialfis: `rnage` -> `range`

src/threading_utils/parallel_for.h
```diff
@@ -0,0 +1,83 @@
...
+   * The i-th thread gets the range [work_range[i], work_range[i+1]). */
+  std::vector<IndexType> work_range = ComputeWorkRange(begin, end, nthread);
+
+  // Launch (nthread - 1) threads, as the main thread should also perform work.
+  std::vector<std::thread> threads;
+  for (std::size_t thread_id = 1; thread_id < nthread; ++thread_id) {
+    threads.emplace_back([&work_range, &func, thread_id]() {
```
trivialfis: Do we need to worry about exception handling?  If so then you might want to use async + future.  Otherwise looks good.

src/CMakeLists.txt
hcho3: Not necessary, since we are already linking `Threads::Threads` target.

src/threading_utils/parallel_for.h
hcho3: Good point. Let me change it to `std::async` instead.

src/CMakeLists.txt
trivialfis: I couldn't remember why I added that to xgboost, maybe for mingw ...  We can think about it when it comes up again.

trivialfis (approved):
LGTM.

PR #311, opened 2021-09-16
Improve vector leaf performance

RAMitchell:
On a large random forest problem, this improves Treelite model generation time from ~650s to 1s.


include/treelite/tree_impl.h
```diff
@@ -737,24 +742,12 @@ template <typename ThresholdType, typename LeafOutputType>
...
-  if (new_end_oft != leaf_vector_.Size()) {
-    throw std::runtime_error("Invariant violated");
-  }
-  std::for_each(&leaf_vector_offset_.at(nid + 1), leaf_vector_offset_.End(),
-                [new_end_oft](std::size_t& x) { x = new_end_oft; });
-
-  Node& node = nodes_.at(nid);
+  leaf_vector_begin_[nid] = begin;
+  leaf_vector_end_[nid] = end;
+  Node &node = nodes_.at(nid);
```
hcho3, line 750: It might be prudent to throw an error if the user attempts to call `SetLeafVector` out-of-order, e.g. `SetLeafVector(nid=1)` followed by `SetLeafVector(nid=0)`. The previous version of the code performs this check. Do you think adding the check will greatly impact performance?

In general, Treelite assumes that the user initializes nodes in the order of node IDs: node 0, node 1, node 2, and so forth; breaking this assumption will cause undefined behavior.

RAMitchell: I don't think there is actually any problem with setting this out of order in the current design. We have a unique map for each nid to the leaf vector, and they are independent, so I don't think the ordering of the leaf vector matters, unless I have missed something. I think the previous throw statement was one of the performance issues unfortunately.

It looks like your python linter has updated and is failing the unchanged python code. What would you like me to do here?

include/treelite/tree_impl.h
```diff
@@ -627,7 +631,8 @@ inline void
 Tree<ThresholdType, LeafOutputType>::Init() {
   num_nodes = 1;
   leaf_vector_.Clear();
-  leaf_vector_offset_.Resize(2, 0);
+  sleaf_vector_begin_.Resize(1, {});
```
hcho3:
```suggestion
  leaf_vector_begin_.Resize(1, {});
```
Stray `s`?

include/treelite/tree_impl.h
hcho3: @RAMitchell Actually you are right. As long as you don't call `SetLeafVector` twice for the same node, things will be okay. How about if we add a check to ensure that `SetLeafVector` gets called at most once for each node ID?

hcho3:
There is some problem with Conda environment. Let me sort that out.

include/treelite/tree_impl.h
RAMitchell: Whoops, I forgot to commit the last changes.

include/treelite/tree_impl.h
RAMitchell: I think this will still not result in any critical failure, there will just be some unused memory in the leaf vector. I can do this change if you want, but I should redo the benchmarks to check if the throw statement kills performance again.

hcho3: Got it. No need to re-run the benchmarks. Let's keep the current logic as it is.

include/treelite/tree_impl.h
hcho3: Any updates on this one?

hcho3: Fixed on my end.

hcho3 (approved):

hcho3:
@RAMitchell I've fixed the CI. We should be able to merge this PR soon.

PR #322, opened 2021-10-28
Support for isolation forests

### LINKED ISSUE #321: Request: predicting for isolation forests
Would be nice to add a prediction formula for isolation forest models, which are meant to produce a standardized output as follows:
```
2^(- avg(leaf scores) / c)
```

With a constant `c` which is pre-calculated for the model (should be supplied as a parameter to the model builder).

david-cortes:
Fixes #321 

This PR adds support for isolation forest model types by:
* Adding a prediction transformation function which follows the formula used by this model type for obtaining a standardized score used for outlier detection.
* Adding extra parameters as needed for this formula to work (a standardizing constant).
* Adding conversors from the scikit-learn implementation.

A small note about the scikit-learn implementation: this PR adds a different prediction type from what scikit-learn outputs when calling `decision_function`. What this PR does is based on the reference paper that introduced this algorithm, and follows the idea of "higher score -> more anomalous" (opposite to scikit-learn, but it's what every other implementation does). Scikit-learn uses this same formula internally to calculate their `decision_function`, but the results will not match with this PR further than a few decimal places because scikit-learn uses a less precise formula than this PR at the moment (explained in https://github.com/scikit-learn/scikit-learn/pull/19087).

I'm not sure if this PR modifies enough files for the documentation to update, and I don't know if the non-python interfaces will work though.


python/treelite/sklearn/__init__.py
```diff
@@ -13,9 +13,10 @@
 from .rf_regressor import SKLRFRegressorMixin
 from .rf_classifier import SKLRFClassifierMixin
 from .rf_multi_classifier import SKLRFMultiClassifierMixin
+from .rf_iforest import SKLiForestMixin
 
 
-def import_model_with_model_builder(sklearn_model):
+def import_model_with_model_builder(sklearn_model):  # pylint: disable=R0911
```
hcho3: The function `import_model_with_model_builder` is for educational purposes only and has been deprecated for other purposes. Let's not add the isolation forest to this function. 
It's sufficient to add support to `import_model`.

david-cortes: Removed.

src/compiler/pred_transform.cc
```diff
@@ -68,6 +70,9 @@ pred_transform_db = {
   - exponential
     Apply the exponential function (exp) element-wise to the margin vector. The
     output will be a vector of length [number of data points].
+  - exponential_standard_ratio
+    Apply the exponential base 2 function (exp2) element-wise to a standardized
+    version of the margin vector. The output will be a vector of length [number of data points].
```
hcho3, line 75:
```suggestion
  - exponential_standard_ratio
    Apply the exponential base 2 function (exp2) element-wise to a standardized
    version of the margin vector. The output will be a vector of length [number of data points].
    Each output element is exp2(-x / c), where x is the margin and c is the standardization constant.
```

hcho3 (changes requested):
LGTM. Thank you for your contribution.

python/treelite/sklearn/__init__.py
hcho3: Thanks!

hcho3:
Thanks!

tzemicheal:
Thanks, @david-cortes  for the contribution. I tested running small inference, couldn't able run the code. Could you provide a small test case running isolation forest in FIL inference? 
```python
from sklearn.ensemble import IsolationForest
import treelite
import treelite.sklearn

X = [[-1.1], [0.3], [0.5], [100]]
clf = IsolationForest(random_state=0).fit(X)
clf.predict([[0.1], [0], [90]])
clf.decision_function([[0.1], [0], [90]])


model = treelite.sklearn.import_model(clf)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-be8734c89694> in <module>
----> 1 model = treelite.sklearn.import_model(clf)

/opt/conda/envs/rapids/lib/python3.7/site-packages/treelite/sklearn/importer.py in import_model(sklearn_model)
    133
    134     if isinstance(sklearn_model, IsolationForest):
--> 135         ratio_c = expected_depth(sklearn_model.max_samples)
    136
    137     node_count = []

/opt/conda/envs/rapids/lib/python3.7/site-packages/treelite/sklearn/importer.py in expected_depth(n_remainder)
     47 def expected_depth(n_remainder):
     48     """Calculates the expected isolation depth for a remainder of uniform points"""
---> 49     if n_remainder <= 1:
     50         return 0
     51     if n_remainder == 2:

TypeError: '<=' not supported between instances of 'str' and 'int'
```

hcho3:
@tzemicheal This PR is not yet part of the stable release of Treelite (2.1.0). Did you install Treelite from the source?

david-cortes:
@tzemicheal Thanks for pointing this out. I've submitted a small PR which should fix it.

tzemicheal:
> @tzemicheal This PR is not yet part of the stable release of Treelite (2.1.0). Did you install Treelite from the source?

Yes, I did install it from the source.

PR #323, opened 2021-10-28
Add parent node ID field to Node

hcho3:
This will come in handy when integrating with SHAP.


hcho3:
This will increase the payload size by 16%. I decided to pursue another way to implement path extraction for GPUTreeSHAP.

PR #325, opened 2021-11-24
When version mismatch happens, show versions

hcho3:
Before:
```
Cannot deserialize model from a different version of Treelite
```

After:
```
Cannot deserialize model from a different version of Treelite.
Currently running Treelite version X.X.X
The model checkpoint was generated from Treelite version X.X.X
```

Also, add the C API function `TreeliteQueryTreeliteVersion` so that users can query the version of the Treelite library.

### DIFF
--- a/include/treelite/tree_impl.h
+++ b/include/treelite/tree_impl.h
@@ -840,7 +840,13 @@ Model::DeserializeTemplate(HeaderPrimitiveFieldHandlerFunc header_primitive_fiel
   header_primitive_field_handler(&minor_ver);
   header_primitive_field_handler(&patch_ver);
   if (major_ver != TREELITE_VER_MAJOR || minor_ver != TREELITE_VER_MINOR) {
-    throw std::runtime_error("Cannot deserialize model from a different version of Treelite");
+    std::ostringstream oss;
+    oss << "Cannot deserialize model from a different version of Treelite." << std::endl
+        << "Currently running Treelite version " << TREELITE_VER_MAJOR << "."
+        << TREELITE_VER_MINOR << "." << TREELITE_VER_PATCH << std::endl
+        << "The model checkpoint was generated from Treelite version " << major_ver << "."
+        << minor_ver << "." << patch_ver;
+    throw std::runtime_error(oss.str());
   }
   header_primitive_field_handler(&threshold_type);
   header_primitive_field_handler(&leaf_output_type);
--- a/src/c_api/c_api_error.cc
+++ b/src/c_api/c_api_error.cc
@@ -10,6 +10,10 @@
 #include <string>
 #include <sstream>
 
+// Stringify an integer macro constant
+#define STR_IMPL_(x) #x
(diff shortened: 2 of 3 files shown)


include/treelite/tree_impl.h
```diff
@@ -840,7 +840,13 @@ Model::DeserializeTemplate(HeaderPrimitiveFieldHandlerFunc header_primitive_fiel
...
   header_primitive_field_handler(&patch_ver);
   if (major_ver != TREELITE_VER_MAJOR || minor_ver != TREELITE_VER_MINOR) {
-    throw std::runtime_error("Cannot deserialize model from a different version of Treelite");
+    std::ostringstream oss;
+    oss << "Cannot deserialize model from a different version of Treelite." << std::endl
+        << "Currently running Treelite version " << TREELITE_VER_MAJOR << "."
+        << TREELITE_VER_MINOR << "." << TREELITE_VER_PATCH << std::endl
+        << "The model checkpoint was generated from Treelite version " << major_ver << "."
+        << minor_ver << "." << patch_ver;
+    throw std::runtime_error(oss.str());
```
levsnv, line 849: This is perfect!

src/c_api/c_api_error.cc
```diff
@@ -25,3 +28,11 @@ const char* TreeliteGetLastError() {
...
   TreeliteAPIErrorStore::Get()->last_error = msg;
 }
+
+const char* TreeliteQueryTreeliteVersion() {
+  std::ostringstream oss;
+  oss << TREELITE_VER_MAJOR << "." << TREELITE_VER_MINOR << "." <<  TREELITE_VER_PATCH;
+  std::string& version_str = TreeliteAPIErrorStore::Get()->version_str;
+  version_str = oss.str();
+  return version_str.c_str();
+}
```
levsnv, line 42: this does solve the case when the user doesn't have a model but wants to check a version (e.g. during automated deployment, or to double-check deployment results). However, it seems this requires to build a separate application, whose sole purpose is to check the treelite version?
I agree that it's a neat API for e.g. server startup-time checking of version. However, it seems to me the same can be achieved by
```C++
static const char[] TREELITE_VERSION = "TREELITE_VERSION_" #TREELITE_VER_MAJOR "." #TREELITE_VER_MINOR "." #TREELITE_VER_PATCH;
```
in https://github.com/dmlc/treelite/blob/2e4169750d79795c5f5e842c2190bb595c236d1d/cmake/version.h.in#L10 ?
The reason I'm insisting on this way is that this allows to look from any linux shell as well as a C/C++ application (the application can strcmp with what it expects, output this into any stream/string/stdout).

hcho3: @levsnv Please see my latest commit. You can now get the version number using `strings`:
```
$ strings libtreelite.dylib -o -t x | grep VERSION
2948ed TREELITE_VERSION_2.1.0
```

levsnv (approved):
looks great! thanks for such a quick turnaround!

PR #330, opened 2021-12-20
Add weighted_n_node_samples field in sklearn importer

hcho3:
Scikit-learn tree models has two fields for sample counts in nodes: `n_node_samples` (unweighted count, int64) and `weighted_n_node_samples` (weighted count, float64). So far, Treelite kept only the unweighted count and discarded the weighted count.

This PR stores `weighted_n_node_samples` in the Treelite object, using the `sum_hess` field.

Required by rapidsai/cuml#4447


divyegala (approved):
Looks great!

PR #332, opened 2022-01-04
Support importing directly from LightGBM model object

### LINKED ISSUE #308: Feature request: treelite.Model.from_lightgbm()
Thank you for this great library.

Is there a way to load a LightGBM model from a booster object (like `treelite.Model.from_xgboost`, but for LightGBM)? If not, would you consider implementing it? My use case does not require saving the model to a file, but I'm forced to do it to be able to load the model into treelite.

hcho3:
Implement `treelite.Model.from_lightgbm` method to import directly from a LightGBM model object.

Closes #308


divyegala (approved):
Looks great!

PR #333, opened 2022-01-08
Fix from_xgboost() for models with categorical splits

hcho3:
Currently, `from_xgboost()` doesn't work if the XGBoost model contains one or more categorical splits. The reason is that the method calls the `save_raw()` method in XGBoost, which uses a legacy serialization method that doesn't support categorical splits.

For `from_xgboost()` to work properly, it needs to access the JSON string serialization of the XGBoost model. Unfortunately, `save_model()` in XGBoost always saves to the disk, and XGBoost doesn't yet have a method to directly serialize the model object as a JSON string in-memory.

As a workaround, `from_xgboost()` calls the internal `__getstate__` method to serialize the model into a JSON string. The `__getstate__` method returns a JSON string of form
```
{
    "Config": { ... }
    "Model": { ... }
}
```
where we need to extract the model object using the `Model` key.

@trivialfis Would you be able to review this? I implemented the workaround since another release of XGBoost isn't yet in the horizon, and this feature is necessary to enable categorical support in cuML TreeExplainer.

### DIFF
--- a/python/treelite/frontend.py
+++ b/python/treelite/frontend.py
@@ -7,6 +7,7 @@ import shutil
 import os
 import json
 from tempfile import TemporaryDirectory
+from pkg_resources import parse_version
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


python/treelite/frontend.py
```diff
@@ -379,6 +379,13 @@ def from_xgboost(cls, booster):
                                 '`xgboost.Booster` object') from e
         if not isinstance(booster, xgboost.Booster):
             raise ValueError('booster must be of type `xgboost.Booster`')
+        if parse_version(xgboost.__version__) >= parse_version('1.0.0'):
+            # TODO(hcho3): Remove this line once XGBoost implements a method to save JSON model
+            # in-memory. (Right now, save_model() always saves to the disk.)
+            model_json_str = booster.__getstate__()['handle'].decode('utf-8')
+            return cls.from_xgboost_json(model_json_str)
```
RAMitchell, line 386: Does 'from_xgboost_json()' get tested for multiple xgboost versions?

RAMitchell (approved):
This PR is fragile to changes to the upstream __getstate__ method, but looks fine for now. This needs to be unblocked. @trivialfis are there plans to add a Python API function for in memory JSON?

trivialfis:
It's added in the ubjson PR as suggested by @hcho3 .  See the additional `raw_format` parameter for `save_raw`.

python/treelite/frontend.py
hcho3: For now it is tested with the latest version.

PR #335, opened 2022-01-12
Remove undefined behavior when predicting with invalid category value

### LINKED ISSUE #328: [GTIL] Remove undefined behavior when negative input is given to a categorical test
https://github.com/dmlc/treelite/blob/2994be4eb58818de8257aca158699b8792d57937/src/gtil/predict.cc#L52

This line is undefined when `fvalue` has a negative value. Casting a negative `float` into `uint32_t` is undefined.

TODO. Explicitly check whether `fvalue` is negative.

hcho3:
Closes #328

@levsnv


src/compiler/ast_native.cc
```diff
@@ -642,6 +642,11 @@ class ASTNativeCompilerImpl {
           "split_index"_a = node->split_index,
           "right_categories_flag"_a = right_categories_flag);
       }
+
+      oss << fmt::format(
+          "((data[{split_index}].fvalue >= 0) && !isinf(data[{split_index}].fvalue) &&"
+          "(fabsf(data[{split_index}].fvalue) <= (float)(1U << FLT_MANT_DIG)) && (",
```
levsnv, line 648: is it necessary to check for infinity when also checking both range boundaries? Looks redundant, because here, a NAN or +-Inf would compare outside the range anyway.

src/gtil/predict.cc
```diff
@@ -49,10 +50,16 @@ inline int NextNodeCategorical(float fvalue, const std::vector<uint32_t>& matchi
   if (std::isnan(fvalue)) {
     return default_child;
   }
-  const auto category_value = static_cast<uint32_t>(fvalue);
-  const bool is_matching_category = (
-      std::find(matching_categories.begin(), matching_categories.end(), category_value)
-      != matching_categories.end());
+  bool is_matching_category;
+  float max_representable_int = static_cast<float>(uint32_t(1) << FLT_MANT_DIG);
+  if (fvalue < 0 || std::fabs(fvalue) > max_representable_int || std::isinf(fvalue)) {
```
levsnv: same here - infinity compares as you'd expect for a normal number. For NAN, one could swap the branches and invert the condition?

tests/python/test_invalid_categocial_input.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
levsnv: nitpicking file name: categocial -> categorical

src/compiler/ast_native.cc
```diff
@@ -642,6 +642,11 @@ class ASTNativeCompilerImpl {
...
+          "split_index"_a = node->split_index);
```
levsnv, line 649: and what do we do with LGBM not conforming? this does look like the ideal behavior, should we just merge this and wait until they start throwing exceptions on invalid categories?

hcho3: @levsnv I think we can persuade them to clarify the behavior for INFs and large floats, since the concern for backward compatibility is lesser.

src/compiler/ast_native.cc
hcho3: NAN has a special handling, since test nodes in Treelite have a default direction (`default_left`). But yes, INF handling can be folded into the range check.

src/compiler/ast_native.cc
levsnv: sounds perfect!

src/gtil/predict.cc
hcho3: > For NAN, one could swap the branches and invert the condition?

Not so easy, since result for NAN depends on the `default_left` flag. It's easier to explicitly check for NAN.

PR #344, opened 2022-01-26
Prototype for multithreaded GTIL: row parallel

hcho3:
Speed up GTIL prediction by running prediction for multiple rows in parallel. This PR exposes the `nthread` parameter to allow the user to control the number of CPU threads to use.

Using https://github.com/triton-inference-server/fil_backend/blob/main/notebooks/categorical-fraud-detection/Fraud_Detection_Example.ipynb:
comparison
where **After** is using all 16 CPU cores. Machine used: c5.4xlarge (16 vCPUs)


src/gtil/predict.cc
```diff
@@ -184,17 +199,23 @@ inline std::size_t PredictImpl(const treelite::ModelImpl<ThresholdType, LeafOutp
 namespace treelite {
 namespace gtil {
 
-std::size_t Predict(const Model* model, const DMatrix* input, float* output, bool pred_transform) {
+std::size_t Predict(const Model* model, const DMatrix* input, float* output, int nthread,
+                    bool pred_transform) {
+  // If nthread <= 0, then use all CPU cores in the system
+  if (nthread <= 0) {
```
trivialfis, line 205: If you are using omp, it might be better to use utilities from omp instead of std thread.

src/gtil/predict.cc
```diff
@@ -161,21 +176,21 @@ inline std::size_t PredictImpl(const treelite::ModelImpl<ThresholdType, LeafOutp
...
       // multi-class classification with gradient boosted trees
       auto output_logic = [task_param](
           const TreeType& tree, int tree_id, int node_id, float* sum) {
         sum[tree_id % task_param.num_class] += tree.LeafValue(node_id);
       };
-      return PredictImplInner(model, input, output, pred_transform, output_logic);
+      return PredictImplInner(model, input, output, nthread, pred_transform, output_logic);
     }
   } else {
     auto output_logic = [task_param](
```
trivialfis: Where is param being used?

hcho3: My bad. Let me remove it from the lambda capture.

src/gtil/predict.cc
hcho3: Fixed

hcho3:
There isn't much difference in performance when using OpenMP:
comparison

So the overhead of provisioning OS threads isn't that big of a deal, it appears.

hcho3:
Merging #355 first, since it produces a lower latency.

PR #354, opened 2022-02-03
Use std:: prefix for size_t and fixed integer types

hcho3:
Consistently use `std::` prefix for `size_t` and other fixed integer types. Also include `<cstddef>` and `<cstdint>` headers.


hcho3:
Will revisit later.

PR #357, opened 2022-02-04
Robust handling of nthread

hcho3:
* Recognize `OMP_THREAD_LIMIT` environment variable. See https://github.com/dmlc/xgboost/pull/7390
* Define `ConfigureThreadConfig()` utility function to robustly sanitize user-provided `nthread` argument.
* Wrap sanitized `nthread` into `ThreadConfig`, to make it clear whether the value is sanitized. Now `ParallelFor()` construct takes in `ThreadConfig` instead of a raw integer `nthread`.

### DIFF
--- a/src/annotator.cc
+++ b/src/annotator.cc
@@ -218,7 +218,7 @@ AnnotateImpl(
   }
 
   // perform reduction on counts
-  for (int tid = 0; tid < thread_config.nthread; ++tid) {
+  for (std::uint32_t tid = 0; tid < thread_config.nthread; ++tid) {
     const size_t off = count_row_ptr[ntree] * tid;
     for (size_t i = 0; i < count_row_ptr[ntree]; ++i) {
       new_counts[i] += counts_tloc[off + i];
--- a/src/annotator.cc
+++ b/src/annotator.cc
@@ -72,7 +72,7 @@ template <typename ElementType, typename ThresholdType, typename LeafOutputType>
 inline void ComputeBranchLoopImpl(
     const treelite::ModelImpl<ThresholdType, LeafOutputType>& model,
     const treelite::DenseDMatrixImpl<ElementType>* dmat, size_t rbegin, size_t rend,
-    ThreadConfig thread_config, const size_t* count_row_ptr, uint64_t* counts_tloc) {
+    const ThreadConfig& thread_config, const size_t* count_row_ptr, uint64_t* counts_tloc) {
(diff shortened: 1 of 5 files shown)


src/annotator.cc
```diff
@@ -69,16 +71,16 @@ void Traverse(const treelite::Tree<ThresholdType, LeafOutputType>& tree,
 template <typename ElementType, typename ThresholdType, typename LeafOutputType>
 inline void ComputeBranchLoopImpl(
     const treelite::ModelImpl<ThresholdType, LeafOutputType>& model,
-    const treelite::DenseDMatrixImpl<ElementType>* dmat, size_t rbegin, size_t rend, int nthread,
-    const size_t* count_row_ptr, uint64_t* counts_tloc) {
-  std::vector<Entry<ElementType>> inst(nthread * dmat->num_col, {-1});
+    const treelite::DenseDMatrixImpl<ElementType>* dmat, size_t rbegin, size_t rend,
+    ThreadConfig thread_config, const size_t* count_row_ptr, uint64_t* counts_tloc) {
```
wphicks: Any reason not to pass `ThreadConfig` by const ref here?

src/threading_utils/parallel_for.h
```diff
@@ -53,8 +54,38 @@ class OMPException {
...
 inline int MaxNumThread() {
-  return omp_get_max_threads();
+  return std::min(std::min(omp_get_num_procs(), omp_get_max_threads()), OmpGetThreadLimit());
+}
+
+/*!
+ * \brief Represent thread configuration, to be used with parallel loops.
+ */
+struct ThreadConfig {
+  int nthread;
```
wphicks: Could we make this a `std::optional<std::size_t>`? We're only using the signed integer for special value signaling, right?

wphicks (approved):
LGTM! Just two small questions, but nothing that should prevent a merge.

src/threading_utils/parallel_for.h
hcho3: Let's use uint32_t actually. It doesn't need to be optional because we always first validate the user input.

src/annotator.cc
hcho3: Not really. Should this be changed to const ref?

hcho3: I'll go ahead and use const ref here. In a future PR, I want to add thread scheduling to `ThreadConfig` struct.

PR #359, opened 2022-02-04
Speed up GTIL prediction with 2D parallel loops

hcho3:
Follow-up to #355

- [x] Improve thread utilization by running parallel predictions with a 2D grid (row_id, tree_id).
- [x] Run performance benchmark.

Stacked on top of #357. See https://github.com/hcho3/treelite/pull/1 to look at the net diff between this PR and #357.

PR #361, opened 2022-02-09
Speed up GTIL prediction with 2D parallel loops (alternative)

hcho3:
Alternative to #359, where two modification were made to reduce latency:
* Combine two OpenMP loops into one
* Allocate memory ahead of time

For the second change, we need to create a Predictor class in GTIL that will hold pre-allocate memory buffer.


hcho3:
Closing in favor of #362

Issue #309, opened 2021-09-13
Use std::span to avoid extra copies


hcho3:
https://github.com/dmlc/treelite/blob/c9766ca66590d31eb20f3ec3125baebdb482487c/include/treelite/tree.h#L412-L421

Using `std::span` here will eliminate extra copies.

Per suggestion of @canonizer

PR #362, opened 2022-02-11
Speed up multithreaded GTIL prediction using tiles

hcho3:
Inspired by XGBoost's CPU predictor. A team of threads now process a tile (block) of data rows. We set the tile size judiciously so that the buffer fits into the smallest CPU cache.

Also:
* Don't use bound-checked `at()` in a tight read loop. Use unsafe `operator[]` instead.
* Remove the use of `std::unordered_map`, as it destroys data locality.
* Specialize for the case where the whole tree consists of only numerical splits.
* Specialize for the case where a data row contains no missing value. (`std::isnan()` is surprisingly costly, as it consists of a floating-point comparison)
* Remove duplicated `std::isnan` check
* Optimize `NextNode()` for the common cases (XGBoost, LightGBM, cuML, sklearn)
* Don't pass right_child to `NextNode()`, taking advantage of the equality `right_child == left_child + 1`.
* In `PredValueByOneTree()`, cache the individual `Node` struct to improve data locality.

These optimizations reduced the number of CPU instructions from 1,286B to 542B in a [microbenchmark](https://github.com/hcho3/gtil_bench).

Using the `large_model` in the [microbenchmark](https://github.com/hcho3/gtil_bench):

|   | Elapsed Time |
| ------------- | ------------- |
| GTIL, status quo (`mainline`)  | 20240 ms  |
| GTIL with tiles  | 5871 ms  |
| **GTIL with tiles + various optimization**  | **2091 ms** |
| XGBoost | 1817 ms |

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -275,12 +275,15 @@ class Tree {
       return cright_;
     }
     inline bool DefaultLeft() const {
+      // Extract the most significant bit (MSB) of sindex_, which encodes the default_left field
       return (sindex_ >> 31U) != 0;
     }
     inline int DefaultChild() const {
+      // Extract the most significant bit (MSB) of sindex_, which encodes the default_left field
       return ((sindex_ >> 31U) != 0) ? cleft_ : cright_;
     }
     inline std::uint32_t SplitIndex() const {
+      // Extract all bits except the most significant bit (MSB) from sindex_.
       return (sindex_ & ((1U << 31U) - 1U));
     }
     inline bool IsLeaf() const {
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -349,14 +349,14 @@ class Tree {
    * \param nid ID of node being queried
    */
   inline int LeftChild(int nid) const {
-    return nodes_.at(nid).cleft_;
+    return nodes_[nid].cleft_;
   }
   /*!
    * \brief index of the node's right child
    * \param nid ID of node being queried
    */
   inline int RightChild(int nid) const {
-    return nodes_.at(nid).cright_;
+    return nodes_[nid].cright_;
   }
   /*!
    * \brief index of the node's "default" child, used when feature is missing
@@ -370,36 +370,36 @@ class Tree {
    * \param nid ID of node being queried
    */
   inline uint32_t SplitIndex(int nid) const {
-    return (nodes_.at(nid).sindex_ & ((1U << 31U) - 1U));
+    return (nodes_[nid].sindex_ & ((1U << 31U) - 1U));
   }
   /*!
    * \brief whether to use the left child node, when the feature in the split condition is missing
    * \param nid ID of node being queried
    */
   inline bool DefaultLeft(int nid) const {
-    return (nodes_.at(nid).sindex_ >> 31U) != 0;
+    return (nodes_[nid].sindex_ >> 31U) != 0;
   }
   /*!
    * \brief whether the node is leaf node
    * \param nid ID of node being queried
    */
   inline bool IsLeaf(int nid) const {
-    return nodes_.at(nid).cleft_ == -1;
+    return nodes_[nid].cleft_ == -1;
   }
   /*!
    * \brief get leaf value of the leaf node
    * \param nid ID of node being queried
    */
   inline LeafOutputType LeafValue(int nid) const {
-    return (nodes_.at(nid).info_).leaf_value;
+    return (nodes_[nid].info_).leaf_value;
   }
   /*!
    * \brief get leaf vector of the leaf node; useful for multi-class random forest classifier
    * \param nid ID of node being queried
    */
   inline std::vector<LeafOutputType> LeafVector(int nid) const {
-    const std::size_t offset_begin = leaf_vector_begin_.at(nid);
-    const std::size_t offset_end = leaf_vector_end_.at(nid);
+    const std::size_t offset_begin = leaf_vector_begin_[nid];
+    const std::size_t offset_end = leaf_vector_end_[nid];
     if (offset_begin >= leaf_vector_.Size() || offset_end > leaf_vector_.Size()) {
       // Return empty vector, to indicate the lack of leaf vector
       return std::vector<LeafOutputType>();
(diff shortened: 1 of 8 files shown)


include/treelite/tree.h
```diff
@@ -326,6 +382,8 @@ class Tree {
   inline void
   DeserializeTemplate(ScalarHandler scalar_handler, ArrayHandler array_handler);
 
+  friend class GTILBridge;  // bridge to enable optimized access to nodes from GTIL
```
trivialfis, line 388: Can this be replaced by a simple getter?

src/gtil/predict.cc
```diff
@@ -19,46 +19,103 @@
...
 
-template <typename T>
-inline int NextNode(float fvalue, T threshold, treelite::Operator op,
-                    int left_child, int right_child, int default_child) {
-  if (std::isnan(fvalue)) {
-    return default_child;
+/*!
+ * \brief A dense feature vector.
+ */
+class FVec {
```
trivialfis, line 43: When I was working on XGB, I find this problematic when data is extremely sparse.

src/gtil/predict.cc
```diff
@@ -70,100 +127,253 @@ inline int NextNodeCategorical(float fvalue, const std::vector<uint32_t>& matchi
...
-                                    const ThreadConfig& thread_config, bool pred_transform,
-                                    OutputFunc output_func) {
-  using TreeType = treelite::Tree<ThresholdType, LeafOutputType>;
-  const size_t num_row = input->GetNumRow();
-  const size_t num_col = input->GetNumCol();
-  std::vector<ThresholdType> row(num_col);
-  const treelite::TaskParam task_param = model.task_param;
-  std::vector<float> sum_tloc(task_param.num_class * thread_config.nthread);
-  std::vector<float> sum_tot(task_param.num_class);
+constexpr std::size_t kBlockOfRowsSize = 64;
```
trivialfis, line 130: Be aware of data sparsity and the number of columns when specifying this parameter, it can take up a lot of memory considering that it's coupled with the number of threads.

src/gtil/predict.cc
```diff
@@ -19,46 +19,103 @@
...
     case treelite::Operator::kEQ:
-      return (fvalue == threshold) ? left_child : right_child;
-    case treelite::Operator::kLT:
-      return (fvalue < threshold) ? left_child : right_child;
-    case treelite::Operator::kLE:
-      return (fvalue <= threshold) ? left_child : right_child;
+      return left_child + !(fvalue == threshold);
     case treelite::Operator::kGT:
-      return (fvalue > threshold) ? left_child : right_child;
+      return left_child + !(fvalue > threshold);
```
trivialfis, line 102: You are making the assumption that nodes are contiguous.  I haven't really verified that for treelite nor xgboost ..  Is it possible that one prunes some nodes then reuse the pruned node indices?

include/treelite/tree.h
```diff
@@ -438,16 +490,16 @@ class Tree {
...
+    const std::size_t offset_end = matching_categories_offset_[nid + 1];
     if (offset_begin >= matching_categories_.Size() || offset_end > matching_categories_.Size()) {
       // Return empty vector, to indicate the lack of any matching categories
       // The node might be a numerical split
-      return std::vector<uint32_t>();
+      return std::vector<std::uint32_t>();
     }
-    return std::vector<uint32_t>(&matching_categories_[offset_begin],
-                                 &matching_categories_[offset_end]);
+    return std::vector<std::uint32_t>(&matching_categories_[offset_begin],
```
trivialfis, line 504: Maybe use a span class as return?

src/gtil/predict.cc
hcho3: In Treelite, the frontend reorders nodes so that they are all contiguous.

hcho3:
@trivialfis The current GTIL is also inefficient for the extremely wide (high dimensional) data. This PR does not make it worse. Creating a predictor that's efficient for the wide data is outside the scope of this PR.

src/gtil/predict.cc
hcho3: The current version of GTIL also suffers from the same problem.

trivialfis:
The code looks good to me.

> The current GTIL is also inefficient for the extremely wide (high dimensional) data.

I'm just worrying that the memory usage might go up significantly in an unexpected way. (n_features * block_size * n_threads) .  I "fixed" the issue in xgboost (found by a user) by specifying the block size to be 1 when density is lower than 0.5.

trivialfis:
I'm fine with the current code if it's not supported.  Just something to be aware of in the future.

hcho3:
@trivialfis Thank you for the detailed review. Let me create a new GitHub issue to note that GTIL is currently inefficient for models trained with very wide data.

include/treelite/tree.h
hcho3: I'm still debating whether I should expose Node class to the end user, since it lacks information pertaining to categorical splits. (You need to query the tree object for that.)

src/gtil/predict.cc
hcho3: Noted.

include/treelite/tree.h
```diff
@@ -264,6 +266,59 @@ class Tree {
...
+
+    /** Getters **/
+    inline int LeftChild() const {
+      return cleft_;
+    }
+    inline int RightChild() const {
+      return cright_;
+    }
+    inline bool DefaultLeft() const {
+      return (sindex_ >> 31U) != 0;
```
wphicks, line 279: Could you add a comment explaining this operation?

include/treelite/tree.h
```diff
@@ -264,6 +266,59 @@ class Tree {
...
+    }
+    inline int DefaultChild() const {
+      return ((sindex_ >> 31U) != 0) ? cleft_ : cright_;
```
wphicks, line 283: Could you also add comments for this and the other bitwise manipulations in here?

wphicks (approved):
LGTM. Requested a couple of additional comments but nothing that should hold up the merge.

include/treelite/tree.h
hcho3: See #309

include/treelite/tree.h
hcho3: Done

PR #365, opened 2022-02-16
Ensure compatibility with XGBoost 1.6+ in from_xgboost() method

hcho3:
@trivialfis Would you be able to take a quick look and see if it's reasonable?

### DIFF
--- a/python/treelite/frontend.py
+++ b/python/treelite/frontend.py
@@ -384,7 +384,7 @@ class Model:
             # For XGBoost version 1.6.0 and later, use save_raw() to export models as JSON string
             model_json_str = booster.save_raw(raw_format="json").decode("utf-8")
             return cls.from_xgboost_json(model_json_str)
-        elif xgb_version >= parse_version("1.0.0"):
+        if xgb_version >= parse_version("1.0.0"):
             # Prior to version 1.6.0, XGBoost offer a method to export models as JSON string
             # in-memory. So use __getstate__ instead.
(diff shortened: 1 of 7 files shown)


hcho3:
Need to think of a way to test with 1.6 dev version of XGBoost.

python/treelite/frontend.py
```diff
@@ -379,10 +379,15 @@ def from_xgboost(cls, booster):
...
         if not isinstance(booster, xgboost.Booster):
             raise ValueError('booster must be of type `xgboost.Booster`')
-        if parse_version(xgboost.__version__) >= parse_version('1.0.0'):
-            # TODO(hcho3): Remove this line once XGBoost implements a method to save JSON model
-            # in-memory. (Right now, save_model() always saves to the disk.)
-            model_json_str = booster.__getstate__()['handle'].decode('utf-8')
+        xgb_version = parse_version(xgboost.__version__)
+        if xgb_version > parse_version("1.5.2"):
+            # For XGBoost version 1.6.0 and later, use save_raw() to export models as JSON string
+            model_json_str = booster.save_raw(raw_format="json").decode("utf-8")
```
trivialfis: Is it necessary to decode it?

trivialfis (commented):
That's much better, we are now using API with stable interface.

python/treelite/frontend.py
hcho3: @trivialfis No it is not necessary. I just need to revise `from_xgboost_json()` to accept a `bytearray` in addition to a string.

PR #380, opened 2022-05-14
[XGBoost 1.6] Handle pseudo-Huber loss with slope

hcho3:
This should fix all integration tests with XGBoost.

### DIFF
--- a/src/frontend/xgboost_json.cc
+++ b/src/frontend/xgboost_json.cc
@@ -8,7 +8,6 @@
 
 #include "xgboost/xgboost_json.h"
 
-#include <fmt/format.h>
 #include <rapidjson/error/en.h>
 #include <rapidjson/document.h>
 #include <rapidjson/filereadstream.h>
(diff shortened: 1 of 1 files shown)


src/frontend/xgboost_json.cc
```diff
@@ -378,7 +377,9 @@ bool ObjectiveHandler::StartObject() {
           push_key_handler<IgnoreHandler>("tweedie_regression_param") ||
           push_key_handler<IgnoreHandler>("softmax_multiclass_param") ||
           push_key_handler<IgnoreHandler>("lambda_rank_param") ||
-          push_key_handler<IgnoreHandler>("aft_loss_param"));
+          push_key_handler<IgnoreHandler>("aft_loss_param") ||
+          push_key_handler<IgnoreHandler>("pseduo_huber_param") ||
+          push_key_handler<IgnoreHandler>("pseudo_huber_param"));
```
trivialfis, line 382: Pushed twice?

hcho3: The one is the mispelled version used in 1.6

trivialfis: Thank you for the fix

PR #413, opened 2022-11-10
[WIP] Implement JSON import

### LINKED ISSUE #408: Implement JSON import
Currently, Treelite offers ability to export its Model objects as JSON strings, but it's not possible to import it back. A third-party libraries (e.g. those from R) may want to emit JSON strings that can later be imported into Treelite.

hcho3:
Closes #408

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -9,6 +9,7 @@
 
 #include <treelite/base.h>
 #include <treelite/version.h>
+#include <treelite/logging.h>
 #include <algorithm>
 #include <map>
 #include <memory>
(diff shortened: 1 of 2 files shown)


hcho3:
Closing this. We should treat JSON as another serializer, with a round-trip guarantee. The JSON serializer should be implemented in terms of `SerializeTemplate` and `DeserializeTemplate`. A proper refactor of serializers is also in order.

The current method `dump_as_json` should be kept separate from the JSON serializer, to provide a more human-readable alternative. XGBoost does this, with `booster.save("xxx.json")` (complete serializer, not human-readable, round-trip supported) and `booster.dump_model(dump_format="json")` (incomplete, human-readable, one-way).

PR #423, opened 2022-12-01
Add allow_unknown_field flag to XGBoost JSON parser

### LINKED ISSUE #416: Allow unknown fields in XGBoost JSON
XGBoost sometimes adds new fields in its JSON serializer. Currently, Treelite throws a fatal error if it encounters an unknown new field. For example, the CI failed in #412 because XGBoost added `boost_from_average` field in the version 1.7.0.

Treelite should be able to gracefully handle extra fields in a XGBoost JSON model.

hcho3:
Closes #416

How `allow_unknown_field` flag is implemented:
* When a handler's `Key()` method is called with an unknown key, it sets its internal state `state_next_field_ignore_=True`.
* The handler will then proceeds to ignore the subsequent value that directly follows the unknown key. This is possible because of the internal state `state_next_field_ignore_`.
* The state `state_next_field_ignore_` gets reset to False, so that the handler will parse other fields as usual.

To enable the `allow_unknown_field` flag, I introduce a new parameter `parser_config` to all C API functions.


hcho3:
@wphicks Can you review this pull request? It implements `allow_unknown_field` flag in the XGBoost JSON parser.

wphicks (approved):
Brilliant! Great implementation and a great feature to have.

hcho3:
@wphicks Thanks for the quick review! Are you okay with the apparent complexity of this solution? I got worried because we are adding new states to handle `allow_unknown_field`.

PR #442, opened 2023-02-18
Implement predict_leaf and predict_per_tree

### LINKED ISSUE #424: [FEA] Return the observations propagated to leaf values
### Background / Motivation
As highlighted in a previous [exploration](https://github.com/linanqiu/treelite-oob/blob/main/treelite_oob.ipynb), identifying which training observations were used to make a prediction can be useful for calculating out-of-bag errors and causal applications. Although the existing Treelite interface can be manipulated to return which observations were involved per tree, other features such as generating a random forest’s weight matrix (see [explanation](https://cdsamii.github.io/cds-demos/forest-weights/forest-weights.html) and a specific [implementation](https://cran.r-project.org/web/packages/Rforestry/Rforestry.pdf) in a library) requires identifying which observations were involved per leaf. This generates useful insights since a single prediction can be thought of a weighted set of observations from the training set.

Although it would be great to implement OOB predictions and weight matrix features natively, it may require non-lossy representation of leaf nodes in the checkpoint format. Instead of just storing the average value of a leaf, the training observations used to calculate the average value would needed as well. An alternative implementation that avoids changing the checkpoint format would be exposing an interface that returns which input observations were propagated to each leaf. Calculating OOB error and a weight matrix would then require a two-pass approach in which the training data is re-propagated to nodes, then compared against new test data handled by the same interface. This approach requires downstream implementation for Treelite users, but is flexible enough to support other applications. For instance, it would be possible to perturb the training data and generate custom weight matrices for other specific insights.

### Potential Interface
Similar to the `pred_margin` parameter, there could be an additional parameter that specifies whether to return the observations associated with each leaf instead of the raw prediction in the `treelite.gtil.predict()` function or in the standard runtime library.

The returned value could remain the same `numpy.ndarray` type, but with some concern regarding memory depending on size of forest. It also requires that during serialization of a tree, each leaf node has a static index. The interface could return a vector with a length equivalent to the number of trees. Each tree index could then hold a mapping of leaf index to the input observation indices propagated to that leaf. This could be represented by a Boolean matrix (number of leaves by number of observations) or a vector of the observations in each leaf. As long as the leaf indices remain consistent across calls, a Treelite user can implement OOB and weight matrix themselves.

### Example Application
Pseudo-code for potential implementation of a weight matrix:

```
training_leaves = predict(treelite_model, training_data, *pred_leaf* =True)
test_leaves = predict(treelite_model, test_data, *pred_leaf* =True)
i, j = number of training observations, number of test observations
weight_matrix = empty matrix of size i x j
for leaf_index, leaf in enumerate(test_leaves):
  for test_obs_index in leaf:
    for training_obs_index in training_leaves[leaf_index]:
       weight_matrix[training_obs_index, test_obs_index] += 1
weight_matrix = rescale weight_matrix to get final representation
```

### LINKED ISSUE #435: [FEA] Interface for fast out-of-bag predictions
### Background
Treelite currently returns a single averaged value across all trees in a forest. However, in some cases such as for out-of-bag predictions, only a subset of the trees may be used for averaging. Out-of-bag predictions can be extremely useful for evaluating models and further model diagnostics. Although Treelite itself does not track which observations were out-of-sample per tree, many libraries do provide this information (e.g. [_generate_unsampled_indices()](https://scikit-optimize.github.io/stable/_modules/sklearn/ensemble/_forest.html) in Sklearn). If Treelite were to expose the individual predictions per tree, users could then calculate out-of-bag results by hand. 

### Potential Implementation
Treelite could return an array of predictions per tree. See this [notebook](https://github.com/linanqiu/treelite-oob/blob/main/treelite_oob.ipynb) for a detailed exploration.  
Alternatively, a user could pre-emptively specify which trees should be used for averaging. If implemented, the existing feature request for [`predict_leaf()`](https://github.com/dmlc/treelite/issues/424) could also be manipulated to return OOB predictions, but would require a much slower multi-pass approach.

hcho3:
* Implement `predict_leaf`. Closes #424
* Implement `predict_per_tree`. Closes #435

Usage:
```python
leaf_pred = treelite.gtil.predict_leaf(model, X_sample)
pred_bytree = treelite.gtil.predict_per_tree(model, X_sample)
```

### DIFF
--- a/docs/treelite-gtil-api.rst
+++ b/docs/treelite-gtil-api.rst
@@ -1,0 +1,12 @@
+=====================================
+General Tree Inference Library (GTIL)
+=====================================
+
+GTIL is a **reference implementation** of a prediction runtime for all Treelite models. It has the following goals:
+
+* **Universal coverage**: GTIL shall support all tree ensemble models that can be represented as Treelite objects.
+* **Accessible code**: GTIL should be written in an easy-to-read style that can be understood to a first-time contributor. We prefer code legibility to performance optimization.
+* **Correct output**: As a reference implementation, GTIL should produce correct prediction outputs.
+
+.. automodule:: treelite.gtil
+   :members:
--- a/include/treelite/c_api.h
+++ b/include/treelite/c_api.h
@@ -474,9 +474,13 @@ TREELITE_DLL int TreeliteFreeModel(ModelHandle handle);
 
 /*!
  * \brief Load a configuration for GTIL predictor from a JSON string.
- * \param config_json a JSON string. Valid entries are:
- *   - nthread (optional): Number of threads used for initializing DMatrix.
- *   - pred_transform (optional, defaults to True): Whether to apply post-prediction transformation
+ * \param config_json a JSON string with the following fields:
+ *   - "nthread" (optional): Number of threads used for initializing DMatrix.
+ *   - "predict_type" (required): Must be one of the following.
+ *     - "default": Sum over trees and apply post-processing
+ *     - "raw": Sum over trees, but don't apply post-processing; get raw margin scores instead.
+ *     - "leaf_id": Output one (integer) leaf ID per tree.
+ *     - "score_per_tree": Output one or more margin scores per tree.
  * \param out Parsed configuration
  * \return 0 for success; -1 for failure
  */
--- a/include/treelite/c_api.h
+++ b/include/treelite/c_api.h
@@ -31,6 +31,8 @@ typedef void* AnnotationHandle;
 typedef void* CompilerHandle;
 /*! \brief handle to a polymorphic value type, used in the model builder API */
 typedef void* ValueHandle;
+/*! \brief handle to a configuration of GTIL predictor */
+typedef void* GTILConfigHandle;
 /*! \} */
 
 /*!
@@ -471,32 +473,63 @@ TREELITE_DLL int TreeliteFreeModel(ModelHandle handle);
  */
 
 /*!
+ * \brief Load a configuration for GTIL predictor from a JSON string.
+ * \param config_json a JSON string. Valid entries are:
+ *   - nthread (optional): Number of threads used for initializing DMatrix.
+ *   - pred_transform (optional, defaults to True): Whether to apply post-prediction transformation
+ * \param out Parsed configuration
+ * \return 0 for success; -1 for failure
+ */
+TREELITE_DLL int TreeliteGTILParseConfig(const char* config_json, GTILConfigHandle* out);
+
+/*!
+ * \brief Delete a GTIL configuration from memory
+ * \param handle Handle to the GTIL configuration to be deleted
+ * \return 0 for success; -1 for failure
+ */
+TREELITE_DLL int TreeliteGTILDeleteConfig(GTILConfigHandle handle);
+
+/*!
+ * \brief Deprecated. Please use \ref TreeliteGTILGetPredictOutputSizeEx instead.
+ */
+TREELITE_DLL int TreeliteGTILGetPredictOutputSize(ModelHandle model, size_t num_row, size_t* out);
+
+/*!
  * \brief Given a batch of data rows, query the necessary size of array to hold predictions for all
  *        data points.
  * \param model Treelite Model object
  * \param num_row Number of rows in the input
+ * \param config Configuration of GTIL predictor. Set this by calling \ref TreeliteGTILParseConfig.
  * \param out Size of output buffer that should be allocated
  * \return 0 for success; -1 for failure
  */
-TREELITE_DLL int TreeliteGTILGetPredictOutputSize(ModelHandle model, size_t num_row, size_t* out);
+TREELITE_DLL int TreeliteGTILGetPredictOutputSizeEx(ModelHandle model, size_t num_row,
+                                                    GTILConfigHandle config, size_t* out);
+
+/*!
+ * \brief Deprecated. Please use \ref TreeliteGTILPredictEx instead.
+ */
+TREELITE_DLL int TreeliteGTILPredict(ModelHandle model, const float* input, size_t num_row,
+                                     float* output, int nthread, int pred_transform,
+                                     size_t* out_result_size);
 
 /*!
  * \brief Predict with a 2D dense array
  * \param model Treelite Model object
  * \param input The 2D data array, laid out in row-major layout
  * \param num_row Number of rows in the data matrix.
- * \param output Pointer to buffer to store the output. Call TreeliteGTILGetPredictOutputSize()
- *               to get the amount of buffer you should allocate for this parameter.
- * \param nthread number of CPU threads to use. Set <= 0 to use all CPU cores.
- * \param pred_transform After computing the prediction score, whether to transform it.
+ * \param output Pointer to buffer to store the output. Call \ref
+ *               TreeliteGTILGetPredictOutputSizeEx to get the amount of buffer you should
+ *               allocate for this parameter.
+ * \param config Configuration of GTIL predictor. Set this by calling \ref TreeliteGTILParseConfig.
  * \param out_result_size Size of output. This could be smaller than
- *                        TreeliteGTILGetPredictOutputSize() but could never be larger than
- *                        TreeliteGTILGetPredictOutputSize().
+ *                        \ref TreeliteGTILGetPredictOutputSizeEx but could never be larger than
+ *                        \ref TreeliteGTILGetPredictOutputSizeEx.
  * \return 0 for success; -1 for failure
  */
-TREELITE_DLL int TreeliteGTILPredict(ModelHandle model, const float* input, size_t num_row,
-                                     float* output, int nthread, int pred_transform,
-                                     size_t* out_result_size);
+TREELITE_DLL int TreeliteGTILPredictEx(ModelHandle model, const float* input, size_t num_row,
+                                       float* output, GTILConfigHandle config,
+                                       size_t* out_result_size);
 
 /*! \} */
 
--- a/include/treelite/c_api.h
+++ b/include/treelite/c_api.h
@@ -476,6 +476,7 @@ TREELITE_DLL int TreeliteFreeModel(ModelHandle handle);
  * \brief Load a configuration for GTIL predictor from a JSON string.
  * \param config_json a JSON string with the following fields:
  *   - "nthread" (optional): Number of threads used for initializing DMatrix.
+ *                           Set <= 0 to use all CPU cores.
  *   - "predict_type" (required): Must be one of the following.
  *     - "default": Sum over trees and apply post-processing
  *     - "raw": Sum over trees, but don't apply post-processing; get raw margin scores instead.
--- a/include/treelite/c_api.h
+++ b/include/treelite/c_api.h
@@ -529,11 +529,16 @@ TREELITE_DLL int TreeliteGTILPredict(ModelHandle model, const float* input, size
  * \param out_result_size Size of output. This could be smaller than
  *                        \ref TreeliteGTILGetPredictOutputSizeEx but could never be larger than
  *                        \ref TreeliteGTILGetPredictOutputSizeEx.
+ * \param out_result_ndim Number of dimensions in the output array.
+ * \param out_result_shape Pointer to an array containing dimensions of the prediction output.
+ *                         This array shall have [out_result_ndim] elements.
+ *                         The product of the elements shall be equal to out_result_size.
  * \return 0 for success; -1 for failure
  */
 TREELITE_DLL int TreeliteGTILPredictEx(ModelHandle model, const float* input, size_t num_row,
                                        float* output, GTILConfigHandle config,
-                                       size_t* out_result_size);
+                                       size_t* out_result_size, size_t* out_result_ndim,
+                                       size_t** out_result_shape);
 
 /*! \} */
 
--- a/python/treelite/gtil/gtil.py
+++ b/python/treelite/gtil/gtil.py
@@ -107,6 +107,7 @@ def predict(
             ctypes.byref(output_size),
         )
     )
+    print(f"Allocating output with size {output_size.value}")
     out_result = np.zeros(shape=output_size.value, dtype=np.float32, order="C")
     out_result_size = ctypes.c_size_t()
     out_result_ndim = ctypes.c_size_t()
--- a/python/treelite/gtil/gtil.py
+++ b/python/treelite/gtil/gtil.py
@@ -3,6 +3,8 @@ General Tree Inference Library (GTIL)
 """
 import ctypes
 import json
+import warnings
+from typing import Optional
 
 import numpy as np
 
@@ -16,10 +18,10 @@ class GTILConfig:
 
     # pylint: disable=too-few-public-methods
 
-    def __init__(self, *, nthread: int, pred_margin: bool):
+    def __init__(self, *, nthread: int, predict_type: str):
         predictor_config = {
             "nthread": nthread,
-            "predict_type": ("raw" if pred_margin else "default"),
+            "predict_type": predict_type,
         }
         predictor_config = json.dumps(predictor_config)
         self.handle = ctypes.c_void_p()
@@ -35,7 +37,12 @@ class GTILConfig:
 
 
 def predict(
-    model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool = False
+        model: Model,
+        data: np.ndarray,
+        *,
+        nthread: int = -1,
+        pred_margin: Optional[bool] = None,
+        predict_type: Optional[str] = None
 ):
     """
     Predict with a Treelite model using General Tree Inference Library (GTIL).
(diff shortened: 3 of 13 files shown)


python/treelite/gtil/gtil.py
```diff
@@ -2,16 +2,50 @@
...
+            "nthread": nthread,
+            "predict_type": predict_type,
+        }
+        predictor_config = json.dumps(predictor_config)
+        self.handle = ctypes.c_void_p()
+        _check_call(
+            _LIB.TreeliteGTILParseConfig(
+                c_str(predictor_config), ctypes.byref(self.handle)
+            )
+        )
```
csadorf: _Note: The diff is somehow messed up, not sure what's going wrong._

I would recommend to implement this class as dataclass to avoid some of the boiler-plate code:
```suggestion
@dataclass
class GTILConfig:
    """Object holding configuration data"""

    predict_type: str
    nthread: int = -1

    def to_json(self):
        return json.dumps(asdict(self))

    def __post_init__(self):
        self.handle = ctypes.c_void_p()
        _check_call(
            _LIB.TreeliteGTILParseConfig(
                c_str(self.to_json()), ctypes.byref(self.handle)
            )
        )
```

python/treelite/gtil/gtil.py
```diff
@@ -2,16 +2,50 @@
...
+            _check_call(_LIB.TreeliteGTILDeleteConfig(self.handle))
+
+
+def predict(
+    model: Model,
+    data: np.ndarray,
+    *,
+    nthread: int = -1,
+    pred_margin: Optional[bool] = None,
+    predict_type: Optional[str] = None
```
csadorf: I would suggest to call this "return_type" or so to make clear that this actually changes the nature of the return value.

python/treelite/gtil/gtil.py
```diff
@@ -2,16 +2,50 @@
...
+):
     """
-    Predict with a Treelite model using General Tree Inference Library (GTIL). GTIL is intended to
-    be a reference implementation. GTIL is also useful in situations where using a C compiler is
-    not feasible.
+    Predict with a Treelite model using General Tree Inference Library (GTIL).
```
csadorf:
```suggestion
    Predict with a Treelite model using the General Tree Inference Library (GTIL).
```

python/treelite/gtil/gtil.py
```diff
@@ -26,34 +60,74 @@ def predict(model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool
...
+            )
+        warnings.warn(
+            "The pred_margin argument is deprecated. Please use "
+            'predict_type="raw" instead.',
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        predict_type = "raw"
+    if not predict_type:
+        predict_type = "default"
```
csadorf: I would check explicitly against `None` to not confuse with falsy values such as _empty string_ or `False`.
Further, I would suggest to capture all valid predict_types in a [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum).

```suggestion
    if predict_type is None:
        predict_type = "default"
    elif predict_type not in list(_PREDICT_TYPES):
        raise ValueError(f"Unknown predict_type: {predict_type}.")
```

python/treelite/gtil/gtil.py
```diff
@@ -26,34 +60,74 @@ def predict(model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool
...
+    if not isinstance(model, Model):
+        raise ValueError('Argument "model" must be a Model type')
+    if (not isinstance(data, np.ndarray)) or len(data.shape) != 2:
+        raise ValueError('Argument "data" must be a 2D NumPy array')
+
+    data = np.array(data, copy=False, dtype=np.float32, order="C")
```
csadorf, line 168: Are we concerned about the potential information loss here?

python/treelite/gtil/gtil.py
```diff
@@ -26,34 +60,74 @@ def predict(model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool
         2D NumPy array, with which to run prediction
     nthread : :py:class:`int <python:int>`, optional
         Number of CPU cores to use in prediction. If <= 0, use all CPU cores.
-    pred_margin : :py:class:`bool <python:bool>`, optional
-        Whether to produce raw margin scores
+    pred_margin : bool, optional
+        Deprecated. Set predict_type="raw" instead.
+    predict_type : string, optional
+        One of the following:
+        * "default": Usual prediction method. Sum over trees and apply post-processing.
```
csadorf: Maybe it is better to also describe what the "default" method does through its name instead of calling it "default". Although I'm a bit at loss what a good name might be.

python/treelite/gtil/gtil.py
```diff
@@ -26,34 +60,74 @@ def predict(model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool
...
+        * "raw": Sum over trees, but don't apply post-processing; get raw margin scores
+                 instead.
+        * "leaf_id": Output one (integer) leaf ID per tree.
+        * "score_per_tree": Output one or more margin scores per tree.
```
csadorf: Without having reviewed the associated implementation yet, would it in principle be possible to obtain the raw, leaf_id, and score_per_tree result types after completing the prediction? In that case we could simply return an object that can be used to obtain all of this data. Something like this:

```python
y = predict(model, X)  # standard approach

# Obtain alternative (rich) result types:
results = predict_results(model, X)
results.prediction
results.raw
results.leaf_id
# and so on ...
```

Using a separate method instead of changing the return type based on a function argument also seems like a slightly better design to me in that case.

The standard predict method is then basically just short for: `y = predict_results(model, X).prediction`.

csadorf (commented):
As requested I only reviewed the Python API for now.

I don't have any major concerns, but I added a few comments and suggestions for potential improvements and alternative approaches. Happy to discuss!

python/treelite/gtil/gtil.py
csadorf: No idea what went wrong with the diff here...

docs/treelite-gtil-api.rst
```diff
@@ -0,0 +1,12 @@
+=====================================
+General Tree Inference Library (GTIL)
+=====================================
+
+GTIL is a **reference implementation** of a prediction runtime for all Treelite models. It has the following goals:
+
+* **Universal coverage**: GTIL shall support all tree ensemble models that can be represented as Treelite objects.
```
wphicks, line 7: Love these clear statements of purpose!

python/treelite/gtil/gtil.py
```diff
@@ -26,34 +60,74 @@ def predict(model: Model, data: np.ndarray, nthread: int = -1, pred_margin: bool
...
-    pred_margin : :py:class:`bool <python:bool>`, optional
-        Whether to produce raw margin scores
+    pred_margin : bool, optional
+        Deprecated. Set predict_type="raw" instead.
+    predict_type : string, optional
+        One of the following:
+        * "default": Usual prediction method. Sum over trees and apply post-processing.
+        * "raw": Sum over trees, but don't apply post-processing; get raw margin scores
+                 instead.
+        * "leaf_id": Output one (integer) leaf ID per tree.
```
wphicks: Can we clarify here what is considered the canonical ID for a leaf? Is it always 0, 1, 2, ... N proceeding breadth-wise across the leaves? If so, does that assume that splits always go "left" on less-than or not-in-category?

include/treelite/c_api.h
```diff
@@ -470,33 +472,74 @@ TREELITE_DLL int TreeliteFreeModel(ModelHandle handle);
...
  * \brief Given a batch of data rows, query the necessary size of array to hold predictions for all
  *        data points.
  * \param model Treelite Model object
  * \param num_row Number of rows in the input
+ * \param config Configuration of GTIL predictor. Set this by calling \ref TreeliteGTILParseConfig.
  * \param out Size of output buffer that should be allocated
  * \return 0 for success; -1 for failure
  */
-TREELITE_DLL int TreeliteGTILGetPredictOutputSize(ModelHandle model, size_t num_row, size_t* out);
+TREELITE_DLL int TreeliteGTILGetPredictOutputSizeEx(ModelHandle model, size_t num_row,
```
csadorf, line 511: I saw that this suffix is also used in other contexts, but could not deduce its meaning – what does `Ex` stand for here? Example?

include/treelite/c_api.h
```diff
@@ -470,33 +472,74 @@ TREELITE_DLL int TreeliteFreeModel(ModelHandle handle);
...
+ * \brief Delete a GTIL configuration from memory
+ * \param handle Handle to the GTIL configuration to be deleted
+ * \return 0 for success; -1 for failure
+ */
+TREELITE_DLL int TreeliteGTILDeleteConfig(GTILConfigHandle handle);
+
+/*!
+ * \brief Deprecated. Please use \ref TreeliteGTILGetPredictOutputSizeEx instead.
+ */
+TREELITE_DLL int TreeliteGTILGetPredictOutputSize(ModelHandle model, size_t num_row, size_t* out);
```
csadorf, line 500: Would it be reasonable and beneficial to avoid the deprecation by simply providing a default value for `config`?

src/c_api/c_api.cc
```diff
@@ -114,6 +118,9 @@ int TreeliteCompilerFree(CompilerHandle handle) {
 }
 
 int TreeliteLoadLightGBMModel(const char* filename, ModelHandle* out) {
+  TREELITE_LOG(WARNING)
+      << "TreeliteLoadLightGBMModel() is deprecated. Please use "
+      << "TreeliteLoadLightGBMModelEx() instead.";
```
csadorf, line 123: Similar to earlier, reasonable and beneficial to avoid the large number of deprecations through a default value?

src/gtil/config.cc
```diff
@@ -0,0 +1,47 @@
...
+      } else if (value == "score_per_tree") {
+        this->pred_type = PredictType::kPredictPerTree;
+      } else {
+        TREELITE_LOG(FATAL) << "Unknown prediction type: " << value;
+      }
+    } else {
+      TREELITE_LOG(FATAL) << "The field \"predict_type\" must be specified";
+    }
+    itr = parsed_config.FindMember("nthread");
+    if (itr != parsed_config.MemberEnd() && itr->value.IsInt()) {
```
csadorf: Should we raise an error in case that the value cannot be interpreted as an integer?

src/gtil/predict.cc
```diff
@@ -37,6 +37,10 @@ namespace {
 using treelite::threading_utils::ThreadConfig;
 using PredTransformFuncType = std::size_t (*) (const treelite::Model&, const float*, float*);
 
+constexpr std::size_t kBlockOfRowsSize = 64;
+constexpr std::int32_t largest_representable_int
+  = (static_cast<std::int32_t>(1) << std::numeric_limits<float>::digits);
```
wphicks: Isn't this the same as the following?
```suggestion
auto static constexpr const largest_representable_int = std::numeric_limits<std::int32_t>::max();
```

src/gtil/predict.cc
```diff
@@ -187,6 +189,41 @@ struct MultiClfProbDistLeafOutputLogic {
   }
 };
 
+struct PredictLeafOutputLogic {
+  template <typename ThresholdType, typename LeafOutputType>
+  inline static void PushOutput(const treelite::Tree<ThresholdType, LeafOutputType>&,
+                                std::size_t, int node_id, float* output, std::size_t) {
+    *output = static_cast<float>(node_id);
```
wphicks, line 194: Am I interpreting this correctly as the overall id of the node within the tree as opposed to the index of this node among leaf nodes?

python/treelite/gtil/gtil.py
wphicks: Reading the implementation, it seems like it's actually an overall node ID among all nodes (not just leaf nodes). I'm trying to understand how that is canonically established when coming from each of the frameworks, though.

wphicks (commented):
Looking good! I had one question about what constitutes the canonical node id, but that may be my own confusion more than anything. In general, though, I don't see any blockers for getting this in.

src/gtil/predict.cc
hcho3: This variable is actually not used. Let me remove it.

src/gtil/predict.cc
hcho3: Yes, we index nodes using all nodes in the tree, including leaf nodes and internal nodes.

wphicks: Okay. In that case, we probably need to be even more explicit about how the ID is defined in the documentation. How do other frameworks handle this? Is it always a breadth-first assignment with a consistent split direction (less than/contained in category list)?

python/treelite/gtil/gtil.py
hcho3: Not yet. If there is a customer request for supporting float64, I will add the support.

python/treelite/gtil/gtil.py
hcho3: I couldn't find a one-word summary of what this does. I'll just leave it as `default`.

python/treelite/gtil/gtil.py
hcho3: No, each mode of prediction has sufficiently different implementation.

src/gtil/predict.cc
hcho3: When reading tree models, Treelite re-assigns new IDs to nodes by traversing each tree breadth-first. So the root node gets ID 0, the left child of the root node gets 1, the right child of the root gets 2, and so forth.
https://github.com/dmlc/treelite/blob/84b4fd31a85bdf9efe95b7e5ee8555a86e723d55/src/frontend/xgboost.cc#L444-L449

Let me go ahead and add a comment about this behavior.

python/treelite/gtil/gtil.py
hcho3: On a second thought, I actually prefer having separate functions: `predict_leaf`, `predict_per_tree`. Let me factor them out into separate functions.

python/treelite/gtil/gtil.py
hcho3: I like how the boilerplate is massively reduced. I applied your suggestion. Thanks!

python/treelite/gtil/gtil.py
hcho3: No longer relevant, as I'm creating separate functions `predict_leaf` and `predict_per_tree`.

python/treelite/gtil/gtil.py
csadorf: I guess I meant that we are silently dropping information as opposed to requiring that data is passed as `float32`.

python/treelite/gtil/gtil.py
```diff
@@ -13,21 +13,29 @@
 from ..util import c_str
 
 
+@dataclass
 class GTILConfig:
     """Object holding configuration data"""
 
-    # pylint: disable=too-few-public-methods
+    predict_type: str
```
csadorf: You could be even more explicit here by using a `typing.Literal`:
```suggestion
    predict_type: Literal["default", "raw", "leaf_id", "score_per_tree"]
```

tests/python/test_gtil.py
```diff
@@ -517,3 +517,167 @@ def test_lightgbm_sparse_categorical_model():
...
+    else:
+        pred_per_tree = treelite.gtil.predict_per_tree(model, X_sample)
+        assert pred_per_tree.shape == (sample_size, num_round)
+        pred = xgb_model.predict(xgb.DMatrix(X_sample), output_margin=True)
+        np.testing.assert_almost_equal(np.sum(pred_per_tree, axis=1), pred, decimal=3)
+
+
+@given(
+    predict_type=sampled_from(["leaf_id", "score_per_tree"]),
+    random_seed=integers(min_value=0, max_value=50),
```
csadorf: I'm a bit confused about explicitly seeding via hypothesis, what is the motivation for this?

hcho3: It was to ensure that (X, y) data pair would come out randomly. Is it better to directly generate them from hypothesis strategies?

include/treelite/c_api.h
hcho3: Default parameters are a C++ feature and are not available for a plain C API.

Context. Treelite exposes a set of functions using a plain C signature, so that they can be called by another language such as Python and Rust. This introduces some limitations:

* Free-standing functions only; no classes or methods.
* No namespaces; use prefixes to avoid name clashes
* No default parameters; all parameters must be explicitly given
* No function overloading.
* Plain types only. C++ objects need to be passed as opaque handles (`void*`).
* No RAII. Objects must be manually allocated and freed.

include/treelite/c_api.h
hcho3: `Ex` stands for Extension.

include/treelite/c_api.h
hcho3: The benefit of exposing a plain C API is that the Treelite Python package does not need Cython. Instead, it uses the `ctypes` module.

src/gtil/config.cc
hcho3: Added

python/treelite/gtil/gtil.py
hcho3: Good point, I'll add a warning when conversion occurs.

python/treelite/gtil/gtil.py
hcho3: Done

tests/python/test_gtil.py
hcho3: I changed the test so that hypothesis directly generates the data matrix.

python/treelite/gtil/gtil.py
```diff
@@ -60,45 +101,73 @@ def predict(
...
     prediction : :py:class:`numpy.ndarray` array
-        Prediction
+        Prediction output. Expected output dimensions:
+
+        - (num_row,) for regressors and binary classifiers
+        - (num_row, num_class) for multi-class classifiers
+
+    Notes
+    -----
+    Treelite assigns a unique integer ID for every node in the tree, including leaf
```
wphicks, line 115: This is great; it clarifies the expected behavior perfectly. The only thing we might add is that this means that predict_leaf output may differ from the training frameworks.

wphicks (approved):

python/treelite/gtil/gtil.py
hcho3: Yes, users should call dump_as_json function to get the correct mapping for IDs.

include/treelite/c_api.h
csadorf: Thanks a lot for the explanation!

PR #449, opened 2023-03-09
Next-generation model builder API, with focus on performance

hcho3:
Create a new module `treelite.experimental.model_builder`, to eventually replace `treelite.ModelBuilder`.

Characteristics
* Fast, because it does not traverse trees while importing
* Only one call to the C API function (whereas the old model builder makes lots of calls to the C layer)
* Use of Python dataclasses makes it intuitive to build models
* The user must order tree nodes in breath-first order. Undefined behavior may result if nodes are ordered differently.

Current limitations
* All nodes currently share the same comparison op. It should be possible to specify separate comparison op for test nodes.
* User should be able to specify `nthread` when creating node objects.

### DIFF
--- a/include/treelite/base.h
+++ b/include/treelite/base.h
@@ -1,5 +1,5 @@
 /*!
- * Copyright (c) 2017-2020 by Contributors
+ * Copyright (c) 2017-2023 by Contributors
  * \file base.h
  * \brief defines configuration macros of Treelite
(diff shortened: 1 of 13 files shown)


hcho3:
Closing this, since it requires too much boilerplate. Users who need good performance may consider using Cython.

PR #506, opened 2023-05-26
Use std::variant in treelite::Model; clean up serializer

### LINKED ISSUE #472: Use std::variant to implement type-based dispatching
Currently, we use class inheritance to implement type-based dispatching . The approach often leads to tons of boilerplate, due to the repetition of the following dispatching logic:
```cpp
void Dispatch(Func func, GenericPtr ptr, TypeInfo underlying_type /* ... */) {
  if (underlying_type == TypeInfo::type_a) {
    func(dynamic_cast<A>(ptr) /* ... */);
  } else if (underlying_type == TypeInfo::type_b) {
    func(dynamic_cast<B>(ptr) /* ... */);
  }  /* ... */
}
```
The boilerplate is worse for `Model::Dispatch`, where dispatching happens with two type variables.

Solution: Use `std::variant<>` instead. It will reduce boilerplate and result into more idiomatic C++.

### LINKED ISSUE #478: Clean up serialization logic
The current logic for serializer is a mess, with feeble attempt at abstraction. In particular, it's unclear what `SerializeTemplate` and `DeserializeTemplate` do, and why two copies of them exist in the header `include/tree/tree.h`.

We need a better abstraction for implementing common logic for PyBuffer and FILE serializers.

Also: we should use `std::ostream` / `std::istream` instead of `FILE*`. This way, we can seamlessly support serialization to in-memory strings.

hcho3:
* Use `std::variant` to implement type dispatching in `treelite::Model`. Previously, type dispatching was implemented with inheritance, leading to non-idiomatic boilerplates. Closes https://github.com/dmlc/treelite/issues/472
* Clean up the serialization logic using mix-in classes. Closes https://github.com/dmlc/treelite/issues/478. Now all the serialization logic lives in `detail/serializer.h`.


hcho3:
Closing in favor of #508

PR #507, opened 2023-05-29
Clean up serialization logic

### LINKED ISSUE #478: Clean up serialization logic

hcho3:
Clean up the serialization logic using mix-in classes. Closes https://github.com/dmlc/treelite/issues/478. Now all the serialization logic lives in `src/serializer.cc`.

I broke up #506 to make it easy to review.


hcho3:
@wphicks Can I get a review for this PR?

include/treelite/detail/serializer_mixins.h
```diff
@@ -0,0 +1,147 @@
...
+  void DeserializeCompositeArray(T* field) {
+    InitArrayFromPyBuffer(field, *it_++);
+  }
+
+  void SkipOptionalField() {
+    ++it_;
+  }
+
+ private:
+  std::vector<PyBufferFrame>::const_iterator it_;
```
wphicks: I'm always a little leery of storing an iterator directly as a struct member. In this case, I think it's okay because we have good guarantees about the lifetime and mutability of the vector relative to the lifetime of the serializer, but is there any other way we could structure this to avoid it?

src/serializer.cc
```diff
@@ -1,67 +1,223 @@
...
+    tree.num_opt_field_per_tree_ = 0;
+    mixin_.SerializePrimitiveField(&tree.num_opt_field_per_tree_);
+
+    /* Extension slot 3: Per-node optional fields -- to be added later */
+    tree.num_opt_field_per_node_ = 0;
+    mixin_.SerializePrimitiveField(&tree.num_opt_field_per_node_);
+  }
+
+ private:
+  MixIn& mixin_;
```
wphicks: Could we use a `shared_ptr` here to avoid worrying about the lifetime of the `MixIn`?

wphicks (approved):
Great refactor! Love the design. Two places where it would be nice to tweak things if we can so that we don't have to worry about lifetime of objects referenced by struct members, but no absolute blockers.

include/treelite/detail/serializer_mixins.h
hcho3: Changed it to store the const reference to `frames`.

src/serializer.cc
hcho3: Done

PR #508, opened 2023-05-31
Use std::variant in treelite::Model

### LINKED ISSUE #472: Use std::variant to implement type-based dispatching

hcho3:
Use `std::variant` to implement type dispatching in `treelite::Model`. Previously, type dispatching was implemented with inheritance, leading to non-idiomatic boilerplates. Closes https://github.com/dmlc/treelite/issues/472

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -26,6 +26,7 @@
 #include <string>
 #include <type_traits>
 #include <utility>
+#include <variant>
 #include <vector>
 
 #define __TREELITE_STR(x) #x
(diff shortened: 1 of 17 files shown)


hcho3:
@wphicks Would you like to take a look? This PR adopts `std::variant<>` in Treelite's model class.

wphicks (approved):

PR #510, opened 2023-06-13
Add additional metadata for trees

hcho3:
The latest XGBoost (dev version) supports two strategies for multi-target models:
* Fit one model per target (fit a group of trees for each target)
* Fit trees that output a vector leaf.
(From https://xgboost.readthedocs.io/en/latest/tutorials/multioutput.html).

It is possible in an XGBoost model to have a mixture of two strategies. For example, to fit a model for three targets, you can fit three (single-output) trees in the first boosting iteration, and fit a single multi-output tree in the second boosting iteration.
To handle this kind of model, **we should clearly annotate each tree to indicate which target/class it belongs to**.

As a corollary, `task_type` is now simpler: there is no more need for distinguishing different types of multiclass classifiers.

Rendered output: https://treelite--510.org.readthedocs.build/en/510/serialization/v4.html

@trivialfis Can you review this revision to the serialization format?

### DIFF
--- a/docs/serialization/v4.rst
+++ b/docs/serialization/v4.rst
@@ -77,7 +77,7 @@ written to the byte sequence in the exact order they appear in the following lis
 
    * Per-tree Metadata
 
-     - ``target_id``: an array of ``int32_t``. ``target[i]`` indicates the target for which the ``i``-th tree produces output.  If the tree is a multi-target tree (i.e. it yields output for all targets), ``target[i]`` is set to -1.
+     - ``target_id``: an array of ``int32_t``. ``target_id[i]`` indicates the target for which the ``i``-th tree produces output.  If the tree is a multi-target tree (i.e. it yields output for all targets), ``target_id[i]`` is set to -1.
        This array is expected to have length ``num_tree``.
      - ``class_id``: an array of ``int32_t``. ``class_id[i]`` indicates the class for which the ``i``-th tree produces output. For vector-leaf trees that produce outputs for multiple classes,
        the corresponding ``class_id[i]`` is set to -1. The ``class_id`` array is expected to have length ``num_tree``.
--- a/docs/serialization/v4.rst
+++ b/docs/serialization/v4.rst
@@ -75,11 +75,14 @@ written to the byte sequence in the exact order they appear in the following lis
      - ``pred_transform``: 256-character long ``char`` array. Stores a human-readable name of the transformation function that's applied to prediction outputs. The unused elements in the array should be padded with null characters (``\0``).
      - ``sigmoid_alpha``: single ``float`` scalar. This model parameter is relevant when ``pred_transform="sigmoid"``.
      - ``ratio_c``: single ``float`` scalar. This model parameter is relevant when ``pred_transform="exponential_standard_ratio"``.
+
    * Per-tree Metadata
+
      - ``target_id``: an array of ``int32_t``. ``target[i]`` indicates the target for which the ``i``-th tree produces output.  If the tree is a multi-target tree (i.e. it yields output for all targets), ``target[i]`` is set to -1.
-        This array is expected to have length ``num_tree``.
+       This array is expected to have length ``num_tree``.
      - ``class_id``: an array of ``int32_t``. ``class_id[i]`` indicates the class for which the ``i``-th tree produces output. For multi-target trees, the corresponding ``class_id[i]`` is set to -1.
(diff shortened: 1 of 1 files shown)


docs/serialization/v4.rst
```diff
@@ -75,6 +75,11 @@ written to the byte sequence in the exact order they appear in the following lis
      - ``pred_transform``: 256-character long ``char`` array. Stores a human-readable name of the transformation function that's applied to prediction outputs. The unused elements in the array should be padded with null characters (``\0``).
      - ``sigmoid_alpha``: single ``float`` scalar. This model parameter is relevant when ``pred_transform="sigmoid"``.
      - ``ratio_c``: single ``float`` scalar. This model parameter is relevant when ``pred_transform="exponential_standard_ratio"``.
+   * Per-tree Metadata
+     - ``target_id``: an array of ``int32_t``. ``target[i]`` indicates the target for which the ``i``-th tree produces output.  If the tree is a multi-target tree (i.e. it yields output for all targets), ``target[i]`` is set to -1.
+        This array is expected to have length ``num_tree``.
+     - ``grove_id``: an array of ``int32_t``. ``grove[i]`` indicates the class for which the ``i``-th tree produces output. If ``task_type == kMultiClfGrovePerClass``, the ``grove_id`` array will have length ``num_tree``;
```
hcho3: Use class_id instead of grove_id.

hcho3:
@RAMitchell @trivialfis @wphicks  I managed to simplify `task_type` (now there's simply `kMultiClf`), by explicitly indicating which target/class each decision tree contributes to. What do you think? I believe the new approach is better, since we make it clear where each tree's output gets accumulated.

docs/serialization/v4.rst
```diff
@@ -21,10 +21,9 @@ We first define a set of enum types to be used in the serialization format.
 
   - ``kBinaryClf`` (0): binary classifier
   - ``kRegressor`` (1): regressor
-  - ``kLearningToRank`` (2): learning-to-rank
-  - ``kMultiClfGrovePerClass`` (3): multi-class classifier with grove-per-class
-  - ``kMultiClfProbDistLeaf`` (4): multi-class classifier with vector leaf outputs producing probability distribution
```
hcho3, line 26: Benefit of new `task_type`: By getting rid of `kMultiClfGrovePerClass` and `kMultiClfProbDistLeaf`, we made the task type orthogonal to the rest of the model specification.

Previously, the task type imposed complex constraints on the model. For example:
https://github.com/dmlc/treelite/blob/3008035e34116e60e0e9a069c20ac6e346979a67/include/treelite/task_type.h#L33-L50

In the new approach, the task type no longer imposes such complex constraints. Instead, the constraints are made explicit via the metadata `target_id` and `class_id`.

trivialfis (commented):
Thank you for working on the specification. It helps to think about how a model should be represented.

My main question regarding the target_id and class_id is whether we need to distinguish them. Alternatively, is it general enough? I find it helpful to think about multi-task where the task is undefined.

For example, XGBoost is working toward quantile regression with confident interval support. This leads to a potential future where we can train a model with a multi-quantile-multi-target model. In such cases, how do we distinguish the task boundary?

One option is by output shape, another one is by explicit annotation like what you are doing with this proposal. I haven't got a clear picture yet, so discussions are welcome!

hcho3:
> My main question regarding the target_id and class_id is whether we need to distinguish them

I gave some thought into this. Multi-class classifiers with grove-per-class semantic is equivalent to a multi-target regressor with model-per-target semantics. For example, the following two models behave quite similarly:

* Six decision trees that produces a scalar output, where Tree 0 and 3 produce the output for target 0, Tree 1 and 4 produce the output for target 1, and Tree 2 and 5 produce the output for target 2.
* Six decision trees that produces a scalar output, where Tree 0 and 3 produce the output for class 0, Tree 1 and 4 produce the output for class 1, and Tree 2 and 5 produce the output for class 2.

That said, I think we should still distinguish between targets and classes. I would like to err on the side of too much details rather than too little. For example, I regret the choice of setting `kBinaryClfRegr` in the task type, since it threw away an important detail (is this model a classifier?).

>  a potential future where we can train a model with a multi-quantile-multi-target model. In such cases, how do we distinguish the task boundary?

I suggest that we use explicit annotation. We can add `task_id` to distinguish the task for different targets. Currently, we assume that all targets have the same task; to remove this restriction, adding `task_id` per target is sufficient.

PR #523, opened 2023-10-20
Improve performance of creating categorical splits

adamreeve:
Creating deep trees with categorical splits can be very slow due to checking and updating the `matching_categories_offset_` values for all nodes after the node being split. This change introduces a separate `matching_categories_size_` array to store the array sizes, and computes any unset offsets in one pass before serialization to keep backwards compatibility of the serialization format, and similarly computes sizes after deserialization.

In a test with some synthetic trees with depth 18, this takes the time to create a tree with the C++ API from 12.8 s to 0.18 s

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -269,7 +269,14 @@ class Tree {
   ContiguousArray<std::size_t> leaf_vector_end_;
   ContiguousArray<std::uint32_t> matching_categories_;
   ContiguousArray<std::size_t> matching_categories_offset_;
+  ContiguousArray<std::size_t> matching_categories_size_;
   bool has_categorical_split_{false};
+
+  // Ensure category offsets are all correct, updating nodes where the offset and size are zero
+  inline void ComputeCategoryOffsets();
+
+  // Compute category sizes based on the offsets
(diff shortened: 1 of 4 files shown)


hcho3:
I am working on a [brand new serialization format](https://treelite.readthedocs.io/en/v4/serialization/v4.html) where category lists are now represented with two arrays `category_list_begin_` and `category_list_end_`. Would this design also meet your performance requirements?

FYI, you can see my work in progress at https://github.com/hcho3/treelite-ng. The logic for setting a categorical test node is [vastly simplified](https://github.com/hcho3/treelite-ng/blob/279cc73d9d2559e88b647e6a2486752ca1203637/include/treelite/detail/tree.h#L137-L155).

adamreeve:
Hi @hcho3, oh excellent, yes it looks like your new design fixes this issue. Is the plan for your new format to be used in Treelite V4? If so then I guess this PR can be closed.

hcho3:
Yes, Treelite 4.0 will use the format documented in https://treelite.readthedocs.io/en/v4/serialization/v4.html

adamreeve:
:+1: right, I missed that that was in the official docs. Thanks, I'll close this.

PR #539, opened 2024-01-11
Don't fetch GTest if it's found by CMake

trxcllnt:
The CMake variable is `GTest_FOUND` not `GTEST_FOUND`.


hcho3:
Thanks! I will make sure to backport this patch to the upcoming 4.0 release.

PR #562, opened 2024-05-16
support NumPy 2.0 (fixes #560)

### LINKED ISSUE #560: NumPy 2.0 support
NumPy 2.0 is coming out soon ( https://github.com/numpy/numpy/issues/24300 ). NumPy 2.0.0rc1 packages for conda & wheels came out 3 weeks back ( https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395 )

To prepare for NumPy 2.0, it might be worthwhile to start testing against NumPy 2 in CI

NumPy has put out [a migration guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide ). More details are in [the release notes]( https://numpy.org/devdocs/release/2.0.0-notes.html )

If Treelite make use of NumPy's C API (and produces wheels that use it), having a release of Treelite with wheels built against NumPy 2.0.0rc1 would be helpful to ensure NumPy 1 & 2 compatible wheels (as wheels built against NumPy 1 won't be compatible with NumPy 2). More details in [this NumPy 2 ABI doc]( https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling )

Also as NumPy is tracking ecosystem support for NumPy 2.0, it would be helpful to share Treelite current support status in issue (with any plans): https://github.com/numpy/numpy/issues/26191

jameslamb:
Fixes #560.

Makes `treelite` compatible with the latest `numpy` release (2.1.0.dev0) and older pre-2.x versions of `numpy`.

As @seberg noted in https://github.com/dmlc/treelite/issues/560#issuecomment-2100670382, the main change here was to change `np.array(..., copy=False)` calls to `np.asarray(...)` calls. As https://numpy.org/doc/stable/reference/generated/numpy.asarray.html notes:

> *No copy is performed if the input is already an ndarray with matching dtype and order. *

Other small changes:

* adds `ruff` to `pre-commit` and its `NPY201` rule to catch use of deprecated or removed NumPy 2.0 code ([as suggested in NumPy's migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin))
  - *this did not catch any other issues, but good protection. to ensure this support keeps working*
* adds `.hypothesis/` directory to `.gitignore`
  - *this was leftover from running the tests locally with `hypothesis` installed*

## How I tested this

I tested `treelite` in a `python:3.10` aarch64 container image.

<details><summary>how I did that (click me)</summary>

Built in a default Python 3.10 image.

```shell
docker run \
    --rm \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it python:3.10 \
    bash
```

Built a wheel.

```shell
apt-get update
apt-get install -y --no-install-recommends \
    build-essential \
    cmake

pip install \
    build

mkdir -p ./dist
python -m build \
    --wheel \
    --outdir ./dist \
    ./python
```

Installed that wheel.

```shell
pip install ./dist/*.whl
```

Installed the latest `numpy` release candidate.

```shell
pip install \
    --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \
    'numpy==2.*,>=0.0.0a'
# 2.1.0.dev0/numpy-2.1.0.dev0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
```

Ran the tests.

```shell
pip install \
    hypothesis \
    lightgbm \
    pytest \
    pytest-cov \
    scikit-learn \
    xgboost

pytest tests/python
```

</details>

Then fixed the tests as seen in the diff here, re-built, re-installed, and re-tested until all the tests were passing.

I also quickly scanned through the things being imported and didn't find any that'd be problematic based on the list of changes at https://numpy.org/devdocs/numpy_2_0_migration_guide.html. The `ruff` plugin should catch those anyway.

```shell
git grep -E 'np\.'
git grep -E 'numpy\.'
```

I'm relying on CI here to test these changes against NumPy 1.x.

### DIFF
--- a/src/c_api/model.cc
+++ b/src/c_api/model.cc
@@ -61,7 +61,7 @@ int TreeliteConcatenateModelObjects(
   API_BEGIN();
   std::vector<treelite::Model const*> model_objs(len, nullptr);
   std::transform(objs, objs + len, model_objs.begin(),
-      [](TreeliteModelHandle e) { return static_cast<const treelite::Model*>(e); });
+      [](TreeliteModelHandle e) { return static_cast<treelite::Model const*>(e); });
   auto concatenated_model = ConcatenateModelObjects(model_objs);
   *out = static_cast<TreeliteModelHandle>(concatenated_model.release());
   API_END();
--- a/tests/python/test_lightgbm_integration.py
+++ b/tests/python/test_lightgbm_integration.py
@@ -237,7 +237,7 @@ def test_lightgbm_sparse_ranking_model(tmpdir):
 
     lgb_model_path = pathlib.Path(tmpdir) / "sparse_ranking_lightgbm.txt"
 
-    dtrain = lgb.Dataset(X, label=y, group=[X.shape[0]])
+    dtrain = lgb.Dataset(X, label=y, group=np.array([X.shape[0]], dtype=np.int32))
     lgb_model = lgb.train(params, dtrain, num_boost_round=1)
     lgb_out = lgb_model.predict(X).reshape((-1, 1, 1))
     lgb_model.save_model(lgb_model_path)
--- a/.github/workflows/coverage-tests.yml
+++ b/.github/workflows/coverage-tests.yml
@@ -9,6 +9,10 @@ on:
 permissions:
   contents: read  # to fetch code (actions/checkout)
 
+defaults:
+  run:
+    shell: bash -l {0}
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
@@ -24,19 +28,18 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: 'true'
-    - uses: mamba-org/provision-with-micromamba@v14
+    - uses: conda-incubator/setup-miniconda@v3
       with:
-        cache-downloads: true
-        cache-env: true
-        environment-name: dev
+        miniforge-variant: Mambaforge
+        miniforge-version: latest
+        activate-environment: dev
         environment-file: ops/conda_env/dev.yml
+        use-mamba: true
     - name: Display Conda env
-      shell: bash -l {0}
       run: |
         conda info
         conda list
     - name: Run tests with test coverage computation
-      shell: bash -l {0}
       run: |
         bash ops/cpp-python-coverage.sh
   win-python-coverage:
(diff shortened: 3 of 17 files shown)


tests/python/test_lightgbm_integration.py
```diff
@@ -237,7 +237,7 @@ def test_lightgbm_sparse_ranking_model(tmpdir):
 
     lgb_model_path = pathlib.Path(tmpdir) / "sparse_ranking_lightgbm.txt"
 
-    dtrain = lgb.Dataset(X, label=y, group=[X.shape[0]])
+    dtrain = lgb.Dataset(X, label=y, group=np.array([X.shape[0]], dtype=np.int32))
```
jameslamb, line 240: This one is `lightgbm`'s responsibility to fix.

Passing a list for `group` ends up invoking `np.array(..., copy=False)`.

```text
    def _list_to_1d_numpy(
        data: Any,
        dtype: "np.typing.DTypeLike",
        name: str
    ) -> np.ndarray:
        """Convert data to numpy 1-D array."""
        if _is_numpy_1d_array(data):
            return _cast_numpy_array_to_dtype(data, dtype)
        elif _is_numpy_column_array(data):
            _log_warning('Converting column-vector to 1d array')
            array = data.ravel()
            return _cast_numpy_array_to_dtype(array, dtype)
        elif _is_1d_list(data):
>           return np.array(data, dtype=dtype, copy=False)
E           ValueError: Unable to avoid copy while creating an array as requested.
E           If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
E           For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

Passing a `numpy` array instead avoids that code path and allows this test to pass on NumPy 2.x

jameslamb: I didn't post enough stacktrace there.

To be clear, that is code inside `lightgbm`: https://github.com/microsoft/LightGBM/blob/e0ac63568c900405e5e438d95c539f3490919aab/python-package/lightgbm/basic.py#L358

src/c_api/model.cc
```diff
@@ -61,7 +61,7 @@ int TreeliteConcatenateModelObjects(
   API_BEGIN();
   std::vector<treelite::Model const*> model_objs(len, nullptr);
   std::transform(objs, objs + len, model_objs.begin(),
-      [](TreeliteModelHandle e) { return static_cast<const treelite::Model*>(e); });
+      [](TreeliteModelHandle e) { return static_cast<treelite::Model const*>(e); });
```
jameslamb: `clang-format` made these style changes to C++ files automatically when I ran

```shell
pre-commit run --all-files
```

tests/python/test_lightgbm_integration.py
seberg: The `int32` is a bit unfortunate, but it, or `intc` is necessary here and seems totally safe to keep indefinitely.

seberg (approved):
This looks all good and safe to me for adapting to NumPy changes (as a NumPy dev).

jakirkham:
There are a couple CI failures, but it looks like something may be misconfigured. For example, am seeing the following [in this CI build]( https://github.com/dmlc/treelite/actions/runs/9115119423/job/25297304352?pr=562 ):

```
Run conda info
/Users/runner/work/_temp/4c3cc175-ed2f-4fc9-bc2e-3f8e90ad7012.sh: line 1: conda: command not found
```

jakirkham:
Thanks Hyunsu! 🙏

jameslamb:
Thanks very much! Once there's a new release of `treelite` with these changes, we can update https://github.com/rapidsai/build-planning/issues/38 and https://github.com/numpy/numpy/issues/26191

PR #578, opened 2024-07-10
Support XGBoost UBJSON

### LINKED ISSUE #573: Add support UBJSON XGBoost models
Starting with XGBoost 2.1.0 release, the default format was changed to UBJSON. It would be best for Treelite to natively support it.

hcho3:
Closes #573

Use https://github.com/nlohmann/json to parse UBJSON as it has the following advantages:

* Idiomatic modern C++
* Available on Conda-forge
* Offers SAX-style parser API, so that we can re-use substantial amount of code from XGBoost JSON parser.


hcho3:
@trivialfis FYI, I managed to implement the UBJSON support without too much work. It reuses the event-based parser we use for parsing XGBoost JSON models.

trivialfis:
Cool, the nlohmann JSON is excellent, glad that it can be directly used here.

PR #582, opened 2024-08-18
Make the lib compilable for gcc 14

dovahcrow:
I'm surprised that gcc11 can compile the lib without proper importing..


hcho3:
Thanks!

PR #587, opened 2024-10-04
Add ability to export as sklearn RF

hcho3:
This PR adds a function to export Treelite models as scikit-learn random forests.
It makes use of the [field accessor API](https://treelite.readthedocs.io/en/latest/treelite-api.html#field-accessors).

Usage:
```python
# tl_model is an treelite.Model object

# Returns either RandomForestClassifier / RandomForestRegressor
clf = treelite.sklearn.export_model(tl_model)  
```

Limitations
* Export to GradientBoosting is not yet supported.
* Treelite models with categorical splits are not yet supported.
* Test coverage currently ensures that `predict()` and `predict_proba()` return correct result. Other estimator metadata may be missing or incorrect.
* Most of export logic is written in the Python layer so may be slow for very large models.

### DIFF
--- a/python/treelite/sklearn/exporter.py
+++ b/python/treelite/sklearn/exporter.py
@@ -3,6 +3,8 @@
 from typing import Any
 
 import numpy as np
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
 
 from ..core import TreeliteError
 from ..model import Model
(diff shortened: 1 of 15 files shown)


src/model_query.cc
```diff
@@ -0,0 +1,54 @@
...
+#include <variant>
+#include <vector>
+
+#include <treelite/tree.h>
+
+namespace {
+
+template <typename ThresholdType, typename LeafOutputType>
+std::uint32_t GetDepth(treelite::Tree<ThresholdType, LeafOutputType> const& tree) {
+  // Visit all trees nodes in breath-first order
```
wphicks: This implementation looks just fine, but mostly for my own edification, is there a reason we do this breadth-first? Generally, when I've done something similar in the past, I just keep track of the maximum depth in a depth-first traversal, which keeps us from having to track the depth of each node separately.

wphicks (approved):
Looks great! One question for my own benefit, but it should not block merge.

src/model_query.cc
hcho3: Not really, I chose breadth-first traversal because I was the most comfortable with it.

> I just keep track of the maximum depth in a depth-first traversal, which keeps us from having to track the depth of each node separately.

Not quite sure about this part. Don't you still have to keep track of the depth of each node somewhere, either in the call stack (if using recursive function calls) or in a `std::stack<std::pair<int, int>>` (if using an explicit stack) ?

python/treelite/sklearn/exporter.py
```diff
@@ -0,0 +1,233 @@
...
+            subestimator_state["n_classes_"] = n_classes[0]
+        else:
+            subestimator_state["n_classes_"] = n_classes.tolist()
+    subestimator.__setstate__(subestimator_state)
+    return subestimator
+
+
+def export_model(model: Model):
+    """
+    Export as scikit-learn RandomForest or GradientBoosting.
```
betatim:
```suggestion
    Export a model as a scikit-learn RandomForest.
```
clean up the gradient boosting bit that isn't supported.

python/treelite/sklearn/exporter.py
```diff
@@ -0,0 +1,233 @@
...
+    # 4. target_id[i] must be either 0 or -1
+    # 5. class_id[i] must be either 0 or -1
+    def raise_not_rf_error():
+        raise NotImplementedError(
+            "Currently only random forests can be exported as scikit-learn model objects. "
+            "Gradient boosting models and other kinds of decision tree models are not yet "
+            "supported."
+        )
+
+    if not average_tree_output:
```
betatim, line 187: For my education, what is this "average tree output"? If it isn't true does that mean the model isn't a random forest? Asking because the message in the exception says that only RFs are supported, so if you can have a RF without `average_tree_output` we probably need a different error message that tells the user that "hey, you are using a RF but average tree output isn't turned on for it"

python/treelite/sklearn/exporter.py
```diff
@@ -0,0 +1,233 @@
...
+        raise_not_rf_error()
+    if not np.all(n_classes == n_classes[0]):
+        raise_not_rf_error()
+    if not np.array_equal(leaf_vector_shape, [n_targets, n_classes.max()]):
```
betatim, line 193: I guess it is the same here? You could have a RF but one of these conditions isn't true, in which case the "only RFs are supported" error message would be confusing.

python/treelite/sklearn/exporter.py
```diff
@@ -0,0 +1,233 @@
...
+        raise_not_rf_error()
+    if not np.all((target_id == 0) | (target_id == -1)):
+        raise_not_rf_error()
+    if not np.all((class_id == 0) | (class_id == -1)):
+        raise_not_rf_error()
+
+    # Heuristics for determining whether the model is a regressor or a classifier
+    if n_classes[0] == 1:
```
betatim: :( that we have to guess this and can't inspect `model` to get a definitive answer

betatim (commented):
Looks reasonable to me. Left a few comments which are mostly for my education.

It could be good to add some tests to trigger the various "this model can't be exported to scikit-learn" conditions. That way it is tested and we make codecov happy.

Could also be worth having one test that uses a round trip from scikit-learn to scikit-learn to check that all the fitted attributes (the ones ending in `_`) are set to the right value (or not set if not supported)

src/model_query.cc
wphicks: Probably easier to show this in code, so here's a [gist](https://gist.github.com/wphicks/9575e83d31ba1feeca8a55c7c9071603). Basically, you can just track the depth and max depth, not the depth of each node separately.

python/treelite/sklearn/exporter.py
hcho3: The precise semantics for `average_tree_output` is given in https://treelite.readthedocs.io/en/latest/serialization/v4.html.

In general, when the error message says "RF", it means "scikit-learn RF." All RF from scikit-learn follows the same set of semantics, such as:
* `average_tree_output` is True: the outputs from individual trees are averaged (not summed) to produce the final output
* The size of leaf output is `(n_targets, n_classes)`.

We can perhaps update the error message as follows: `this Treelite model cannot be represented as scikit-learn random forest`.

python/treelite/sklearn/exporter.py
hcho3: See my remark above. For all scikit-learn RF, this condition holds true.

python/treelite/sklearn/exporter.py
hcho3: It turns out we can query the model type directly: `header_accessor.get_field("task_type")`

betatim: Cool! Naively that seems like "the right thing to do" (tm) then no?

hcho3: Changed it to query the model type directly.

hcho3:
Created #588 to remind myself to remove the pin `cpplint==1.6.1` later.

PR #600, opened 2025-01-07
tree.h: Fix SetModelPresetVariant variant_index template parameter

### LINKED ISSUE #599: Incorrect type for SetModelPresetVariant
See https://github.com/dmlc/treelite/blob/d0eef6254e2ddc39d2f886a4e2226cf1d1676eb3/include/treelite/tree.h#L431

The template parameter `variant_index` type is "int", while it should be "std::size_t" for later use in `std::variant_alternative_t<variant_index, ModelPresetVariant>;`

See https://en.cppreference.com/w/cpp/utility/variant/variant_alternative

This fixes compilation on very strict compiler settings.

> treelite/include/treelite/tree.h:436:88: error: conversion to ‘long unsigned int’ from ‘int’ may change the sign of the result [-Werror=sign-conversion]
  436 |       using ModelPresetT = std::variant_alternative_t<variant_index, ModelPresetVariant>;

davidgraeff:
Fixes compiling with -Werror=sign-conversion
Fixes #599

### DIFF
--- a/include/treelite/tree.h
+++ b/include/treelite/tree.h
@@ -429,7 +429,7 @@ class ModelPreset {
 using ModelPresetVariant = std::variant<ModelPreset<float, float>, ModelPreset<double, double>>;
 
 template <std::size_t variant_index>
-ModelPresetVariant SetModelPresetVariant(int target_variant_index) {
+ModelPresetVariant SetModelPresetVariant(std::size_t target_variant_index) {
   ModelPresetVariant result;
   if constexpr (variant_index != std::variant_size_v<ModelPresetVariant>) {
     if (variant_index == target_variant_index) {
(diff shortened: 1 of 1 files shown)


include/treelite/tree.h
```diff
@@ -428,7 +428,7 @@ class ModelPreset {
 
 using ModelPresetVariant = std::variant<ModelPreset<float, float>, ModelPreset<double, double>>;
 
-template <int variant_index>
+template <std::size_t variant_index>
 ModelPresetVariant SetModelPresetVariant(int target_variant_index) {
```
hcho3:
```suggestion
ModelPresetVariant SetModelPresetVariant(std::size_t target_variant_index) {
```

hcho3 (changes requested):

include/treelite/tree.h
hcho3: The function argument should also be `std::size_t`.

hcho3 (approved):

PR #620, opened 2025-08-17
fix: iForest feature mapping

willy-liu:
This PR fixes the feature mapping logic when importing Isolation Forest models.

### Bug Description
The issue occurs when max_features is not equal to the number of input features, e.g.:
```
clf = IsolationForest(max_features=0.5)
clf = IsolationForest(max_features=10)
```

### Changes:
- Use `enumerate` to track tree index in `estimators_`.
- For Isolation Forest:
  - Handle feature subsampling when `max_features != 1.0` using
    `estimators_features_[tree_idx]`.
  - Add proper feature indices instead of raw `tree.feature`.
- For Other Trees:
  - Move `feature.add` to else section.
- Updated comments for clarity.

This ensures feature indices are correctly aligned with the training subsample,
avoiding incorrect feature mappings in downstream usage.

### DIFF
--- a/tests/python/test_sklearn_integration.py
+++ b/tests/python/test_sklearn_integration.py
@@ -194,9 +194,10 @@ def test_skl_converter_iforest(dataset):
     out_pred = treelite.gtil.predict(tl_model, X)
     np.testing.assert_almost_equal(out_pred, expected_pred)
 
+
 @given(
     dataset=standard_regression_datasets(),
-    max_feat=floats(min_value=0.2, max_value=0.8)
+    max_feat=floats(min_value=0.2, max_value=0.8),
 )
 @settings(**standard_settings())
 def test_skl_converter_iforest_feature_subsampling(dataset, max_feat):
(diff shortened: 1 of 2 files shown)


hcho3:
@willy-liu Can you add a test to cover the case with `max_features != 1.0` ?

tests/python/test_sklearn_integration.py
```diff
@@ -194,6 +194,28 @@ def test_skl_converter_iforest(dataset):
...
+    dataset=standard_regression_datasets(),
+    max_feat=floats(min_value=0.2, max_value=0.8)
+)
+@settings(**standard_settings())
+def test_skl_converter_iforest_feature_subsampling(dataset, max_feat):
+    """Scikit-learn isolation forest with feature subsampling"""
+    X, _ = dataset
+    clf = IsolationForest(
+        max_samples=64,
+        max_features=max_feat,
```
willy-liu, line 208: Added test case where max_features is randomly sampled in range [0.2, 0.8]

willy-liu:
@hcho3 Please review the test, Thank you!

tests/python/test_sklearn_integration.py
willy-liu: This test fails on mainline, but passes with this PR, verifying the fix works as intended.

PR #651, opened 2025-12-19
Optimize Scikit-learn model loading by adding Bulk Tree Construction API 

dantegd:
This PR introduces a bulk tree construction API that significantly improves performance when importing scikit-learn RandomForest models into Treelite. In my benchmarks, the new API achieves ~7-10x speedup over the existing node-by-node construction approach of the current sklearn loader. 

The current implementation spends significant time in per-node overhead due to:

- Repeated ModelBuilder method calls for each node
- Python-C++ boundary crossing overhead accumulating across millions of nodes
- Memory allocation patterns that don't benefit from bulk operations

This becomes a bottleneck in workflows like cuML's `RandomForestClassifier.from_sklearn()`, where treelite import time dominates the conversion process.

This PR implements a `BulkConstructTree` friend function that directly populates the Tree class's internal ContiguousArray members in a single pass, bypassing the `ModelBuilder` abstraction for sklearn imports.

Initial benchmarks: 

Configuration | Total Nodes | Old API (ms) | Bulk API (ms) | Speedup
-- | -- | -- | -- | --
classifier, 50 trees, depth=10 | 39,844 | 13.5 | 1.8 | 7.45x
classifier, 100 trees, depth=15 | 351,826 | 77.3 | 10.2 | 7.54x
classifier, 300 trees, depth=20 | 2,520,062 | 544.9 | 60.7 | 8.98x
regressor, 100 trees, depth=15 | 978,436 | 195.6 | 18.8 | 10.42x

</div><div role="presentation" aria-hidden="true" class="invisible scrollbar horizontal" style="visibility: hidden; opacity: 0; pointer-events: none; position: absolute; width: 800px; height: 10px; left: 0px; bottom: 0px;"><div class="slider" style="background: 0% 0% repeat rgba(121, 121, 121, 0.4); position: absolute; top: 0px; left: 0px; height: 10px; transform: translate3d(0px, 0px, 0px); contain: strict; width: 800px;"></div></div><div role="presentation" aria-hidden="true" class="invisible scrollbar vertical" style="visibility: hidden; opacity: 0; pointer-events: none; position: absolute; width: 0px; height: 164px; right: 0px; top: 0px;"><div class="slider" style="background: 0% 0% repeat rgba(121, 121, 121, 0.4); position: absolute; top: 0px; left: 0px; width: 10px; transform: translate3d(0px, 0px, 0px); contain: strict; height: 164px;"></div></div></div><br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,16 +10,16 @@ repos:
   - id: check-added-large-files
     args: ["--maxkb=4000"]
 - repo: https://github.com/psf/black
-  rev: 25.12.0
+  rev: 26.1.0
   hooks:
   - id: black
 - repo: https://github.com/pycqa/isort
-  rev: 7.0.0
+  rev: 8.0.0
   hooks:
   - id: isort
     args: ["--profile", "black", "--filter-files"]
 - repo: https://github.com/MarcoGorelli/cython-lint
-  rev: v0.18.1
+  rev: v0.19.0
   hooks:
   - id: cython-lint
   - id: double-quote-cython-strings
@@ -49,12 +49,12 @@ repos:
       additional_dependencies: [cpplint==1.6.1]
       types_or: [c++]
 - repo: https://github.com/pre-commit/mirrors-mypy
-  rev: v1.19.0
+  rev: v1.19.1
   hooks:
     - id: mypy
       additional_dependencies: [types-setuptools]
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.14.8
+  rev: v0.15.2
   hooks:
     - id: ruff
       args: ["--config", "python/pyproject.toml"]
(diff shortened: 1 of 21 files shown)


hcho3 (changes requested):
Can we go ahead and simply remove the old sklearn model builder functions? So `LoadSKLearnRandomForestRegressorBulk` should be simply called `LoadSKLearnRandomForestRegressor`, etc.

I don't see a good reason to keep the old functions around, if the new functions are equivalent in functionalities but faster.

hcho3 (approved):

PR #655, opened 2026-02-24
Compatibility fix for XGBoost 3.2

chyunsu3:
Since dmlc/xgboost#11814, XGBoost stores leaf weights as a separate field for multi-target trees.

Since multi-target trees are an experimental feature, we do not provide a compatibility layer for older versions of multi-target trees.


hcho3:
@trivialfis Can you help review? This updates Treelite to support new layout for multi-target trees.

trivialfis (approved):
The change looks good. What do you think of the new layout? Theoretically, the leaf weight can be omitted if reduced gradient is not used. But making an optional field makes the code and more complex.

hcho3:
@trivialfis Is reduced gradient only applicable for multi-target trees? If so, I like the layout.

trivialfis:
Yes, vector leaf only

PR #671, opened 2026-07-13
Migrate python build to scikit build core

### LINKED ISSUE #666: Migrate Treelite to scikit-build-core
Should be analogous to https://github.com/dmlc/xgboost/pull/12219. Migrating to scikit-build-core will allow us to remove the custom backend code in `python/packager`.

CavRiley:
Migrates from the hatch Python build backend to scikit-build-core, a build backend for creating Python modules with cmake.

Follows similarly to the xgboost migration to scikit-build-core seen [here](https://github.com/dmlc/xgboost/pull/12219) 

Closes #666


.gitignore
```diff
@@ -44,9 +44,11 @@
 /python/dist/
 /python/build/
 /python/treelite.egg-info/
+python/cpp_src/
 /runtime/python/dist/
 /runtime/python/build/
 /runtime/python/treelite_runtime.egg-info/
+/runtime/python/cpp_src/
```
chyunsu3: We no longer have a `runtime/` subdirectory. Let's remove these rules.

.github/workflows/misc-tests.yml
```diff
@@ -112,11 +112,10 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=/opt/treelite
+        cmake .. -GNinja -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DCMAKE_INSTALL_LIBDIR=lib
         ninja install -v
         cd ../python
-        pip install --force-reinstall -v . --config-settings use_system_libtreelite=True \
-          --config-settings system_libtreelite_dir=/opt/treelite/lib
```
chyunsu3, line 119: We are removing the ability to configure a custom path for system libtreelite. (That's why the test is titled `Test Treelite with custom libpath`.) Any way to restore this feature?

chyunsu3 (changes requested):

.github/workflows/misc-tests.yml
CavRiley: I restored it. When `TREELITE_SYSTEM_LIBTREELITE_DIR` is passed (with `TREELITE_USE_SYSTEM_LIBTREELITE`), CMake configures a path_config.py template with that path included and installs it into the wheel. This is similar to before but done in Cmake rather than the old build hook. (I also renamed the flag `USE_SYSTEM_LIBTREELITE` to `TREELITE_USE_SYSTEM_LIBTREELITE` to namespace the CMake define)

chyunsu3 (approved):

PR #684, opened 2026-08-18
Export more metadata from `export_model()`

chyunsu3:
* Export `n_node_samples` and `weighted_n_node_samples` fields when exporting Treelite model as a scikit-learn model.
* Use correct estimator classes when exporting isolation forests.
* Add test coverage for round-trip.
* Fix a bug in `data_count()` setter in the model builder API.
* Persist `offset_` in the Treelite model object.

### DIFF
--- a/python/treelite/sklearn/exporter.py
+++ b/python/treelite/sklearn/exporter.py
@@ -103,8 +103,25 @@ def _export_tree(
     nodes["feature"] = tree_accessor.get_field("split_index")
     nodes["threshold"] = tree_accessor.get_field("threshold")
     nodes["impurity"] = np.nan
-    nodes["n_node_samples"] = -1
-    nodes["weighted_n_node_samples"] = np.nan
+    data_count = tree_accessor.get_field("data_count").astype(np.intp)
+    data_count_mask = tree_accessor.get_field("data_count_present").astype(np.bool)
+    if data_count.size == 0:
+        nodes["n_node_samples"] = np.full((n_nodes,), fill_value=-1, dtype=np.intp)
+    else:
+        data_count[~data_count_mask] = -1
+        nodes["n_node_samples"] = data_count
+    # TODO(chyunsu3): Rename field sum_hess -> weighted_data_count
+    weighted_data_count = tree_accessor.get_field("sum_hess").astype(np.float64)
+    weighted_data_count_mask = tree_accessor.get_field("sum_hess_present").astype(
+        np.bool
+    )
+    if weighted_data_count.size == 0:
+        nodes["weighted_n_node_samples"] = np.full(
+            (n_nodes,), fill_value=np.nan, dtype=np.float64
+        )
+    else:
+        weighted_data_count[~weighted_data_count_mask] = np.nan
+        nodes["weighted_n_node_samples"] = weighted_data_count
     nodes["missing_go_to_left"] = tree_accessor.get_field("default_left")
 
     if n_targets == 1 and n_classes[0] == 1:
@@ -175,8 +192,16 @@ def export_model(model: Model) -> Any:
     # pylint: disable=too-many-locals
     try:
         from sklearn import __version__ as sklearn_version
-        from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
-        from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
+        from sklearn.ensemble import (
+            IsolationForest,
+            RandomForestClassifier,
+            RandomForestRegressor,
+        )
+        from sklearn.tree import (
+            DecisionTreeClassifier,
+            DecisionTreeRegressor,
+            ExtraTreeRegressor,
+        )
     except ImportError as e:
         raise TreeliteError("This function requires scikit-learn package") from e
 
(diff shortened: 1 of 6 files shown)


chyunsu3:
@JulienAu Can you review this pull request?

JulienAu:
Happy to, taking a look now.

JulienAu (commented):
@chyunsu3 Thanks for the quick turnaround on this. I reviewed the diff and ran the branch's exporter against the compiled treelite 4.7.0 runtime (the changed files are pure Python), on both sklearn round trips and cuML model bytes.

The `TreeliteModelBuilderDataCount` fix is correct, and the sklearn round trip works as advertised: data_count populated (root equals max_samples_), correct estimator classes, n_node_samples preserved exactly.

Two things worth considering:

1. The exported IsolationForest cannot score: only `estimators_`, `n_outputs_`, `n_features_in_` and `_sklearn_version` are set, so `score_samples` raises `AttributeError: '_max_features'`. Most of the missing state is derivable from the trees, and I verified that populating it plus the default `offset_ = -0.5` gives exact score parity on the round trip. `offset_` itself is not recoverable from the serialized model (base_scores stays 0), so contamination-fitted models would predict differently. Whichever contract you prefer (populate with a documented -0.5 assumption, or keep the export structural with an explicit error on scoring), a `score_samples` call in test_iforest_round_trip would pin it down.

2. Present-but-zero data_count flows through silently. cuML's isolation forest export currently emits exactly that (checked on nightly bytes: every node present with value 0, task_type kRegressor), which would export as n_node_samples equal to 0 everywhere and break scoring downstream. Since a reachable node can never hold zero samples, failing loudly on non-positive present values seems safer. It also means NVIDIA/cuml#8483 can only adopt `exported.estimators_` once the cuML C++ export writes real counts and tags kIsolationForest; happy to coordinate that follow-up.

Minor: `np.bool` is missing on numpy 1.24 through 1.26 and the package floor allows them; `np.bool_` works everywhere.

chyunsu3:
@JulienAu Can you take another look? 

1. Now the exported IsolationForest can score. Attributes including `_max_samples`, `_max_features`, and `estimators_features_` are now exported. Importantly, the `offset_` value is now stored as part of the Treelite model object, so you can now recover `offset_` from the serialized model.

```python
offset = treelite_model.attributes.get("sklearn_iforest_offset", -0.5)
```

2. As for the zero `data_count`, I think cuML should be updated to export the correct `data_count` and `data_count_present` when serializing cuML IsolationForest as Treelite.

JulienAu (commented):
Took another look.

Storing offset_ as a model attribute is a nice solution, I confirmed it survives serialize_bytes and deserialize_bytes, which is the path cuML will use. Scoring parity is exact on my side for the default and contamination cases.

One real issue: with bootstrap=True the scores drift (about 1e-3 in my runs). sklearn fits the bagged trees with sample weights instead of repeating rows, so the root n_node_samples counts distinct rows (63 in my run) and the derived _max_samples comes out wrong. The root weighted_n_node_samples is exactly max_samples on every tree, so I put the fix as a suggestion: with it I get exact parity in every configuration I tested (bootstrap on and off, with and without user sample_weight).

Small note: the public max_samples_ attribute is not set. Scoring only needs the private one, but users may look for it.

python/treelite/sklearn/exporter.py
```diff
@@ -266,6 +296,38 @@ def raise_not_rf_error(reason):
...
+        except KeyError:
+            warnings.warn(
+                "Treelite model does not store attribute 'sklearn_iforest_offset'; "
+                "setting it to the default value of -0.5...",
+                UserWarning,
+            )
+            offset = -0.5
+
+        # Compute max_samples by computing max over n_node_samples from each tree
+        max_samples = max(estimator.tree_.n_node_samples[0] for estimator in estimators)
```
JulienAu:
```suggestion
        # Compute max_samples by taking the max over the weighted root counts
        # (with bootstrap=True the unweighted root only counts distinct rows)
        max_samples = int(
            max(estimator.tree_.weighted_n_node_samples[0] for estimator in estimators)
        )
```

python/treelite/sklearn/importer.py
```diff
@@ -79,6 +80,18 @@ def import_model(sklearn_model) -> Model:
...
         # tl_model is a Treelite representation of clf
 
+    To reproduce the output of :py:meth:`~sklearn.ensemble.IsolationForest.decision_function`,
+    retrieve the value of :py:attr:`~sklearn.ensemble.IsolationForest.offset_` and apply it,
+    as follows:
+
+    .. code-block:: python
+
+        # Treelite model stores an optional list of attributes (as a JSON string).
+        # We can retieve `offset_` from it.
```
JulienAu:
```suggestion
        # We can retrieve `offset_` from it.
```

JulienAu:
Forgot to answer your data_count point: that matches what csadorf already scoped in NVIDIA/cuml#8420, populate data_count in the Treelite export, then simplify the Python side.

chyunsu3:
@JulienAu Can you look at it one last time? I addressed all your feedback.

JulienAu (approved):
Ran my test bench once more on the latest head: exact parity everywhere including bootstrap, offset recovery survives serialize_bytes, and max_samples_ is set. Looks good to me, thanks for the quick turnaround.

PR #694, opened 2026-08-26
[CI] Various CI fixes

chyunsu3:
* Fix pre-commit errors:
> python/treelite/sklearn/importer.py:434: error: Unexpected keyword argument "start" for overloaded function "arange"  [call-overload]
tests/python/util.py:101: error: Value of type "int" is not indexable  [index]
python/treelite/frontend.py:199: error: Module has no attribute "Booster"  [attr-defined]
python/treelite/frontend.py:201: error: Module has no attribute "`__version__`"  [attr-defined]

* Pin `lcov<2.0` to work around error:
> lcov: ERROR: (inconsistent) mismatched end line for _ZN8treelite34SerializerRoundTrip_TreeStump_Test8TestBodyEv at /home/runner/work/treelite/treelite/tests/cpp/test_serializer.cc:163: 163 -> 170 while capturing from ./build/tests/cpp/CMakeFiles/treelite_cpp_test.dir/test_serializer.cc.gcda

### DIFF
--- a/python/treelite/sklearn/importer.py
+++ b/python/treelite/sklearn/importer.py
@@ -431,9 +431,7 @@ def _import_hist_gradient_boosting(sklearn_model) -> Model:
                 feat_remapper[n_categorical + num_idx] = i
                 num_idx += 1
     else:
-        feat_remapper = np.arange(
-            start=0, stop=sklearn_model.n_features_in_, dtype=np.int32
-        )
+        feat_remapper = np.arange(0, stop=sklearn_model.n_features_in_, dtype=np.int32)
 
     n_categorical_splits = known_cat_bitsets.shape[0]
     n_trees = 0
--- a/tests/python/util.py
+++ b/tests/python/util.py
@@ -94,7 +94,9 @@ def to_categorical(
     cat_cols = (cat_cols * rough_n_categories).astype(int)
 
     # Mix categorical and numerical columns in a random order
-    new_col_idx = rng.choice(n_features, n_features, replace=False, shuffle=True)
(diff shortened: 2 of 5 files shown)


tests/python/util.py
```diff
@@ -94,7 +94,9 @@ def to_categorical(
     cat_cols = (cat_cols * rough_n_categories).astype(int)
 
     # Mix categorical and numerical columns in a random order
-    new_col_idx = rng.choice(n_features, n_features, replace=False, shuffle=True)
+    new_col_idx = np.asarray(
+        rng.choice(n_features, n_features, replace=False, shuffle=True)
+    )
```
chyunsu3, line 99: `mypy` is conservative and assuming `int` return type for `rng.choice()`. We explicitly annotate the return type (`np.ndarray`) here.

python/treelite/sklearn/importer.py
```diff
@@ -431,9 +431,7 @@ def _import_hist_gradient_boosting(sklearn_model) -> Model:
                 feat_remapper[n_categorical + num_idx] = i
                 num_idx += 1
     else:
-        feat_remapper = np.arange(
-            start=0, stop=sklearn_model.n_features_in_, dtype=np.int32
-        )
+        feat_remapper = np.arange(0, stop=sklearn_model.n_features_in_, dtype=np.int32)
```
chyunsu3, line 434: The `start` parameter in `np.arange` is now a positional-only argument. See https://github.com/numpy/numpy/issues/30368#issuecomment-3609154302

PR #705, opened 2026-09-02
statically link GTest, update to 1.18.0, fetch it with CPM

### LINKED ISSUE #704: [c++] test executables are not relocatable
## Description

Consider the following:

```shell
cmake -B build -S . -DBUILD_SHARED_LIBS=ON -DBUILD_CPP_TEST=ON
```

That `BUILD_SHARED_LIBS=ON` tells CMake to build a dynamic library (`libtreelite.so`) but also leaks through to other targets, and results in dynamic linkage to GoogleTest for the test executable.

```console
$ ldd ./build/treelite_cpp_test  | grep gtest
        libgtest.so.1.14.0 => /home/jlamb/repos/treelite/build/lib/libgtest.so.1.14.0 (0x000072eba32bc000)
```

This means that to run that executable, you have to either preserve the build directory or provide that version of GoogleTest some other way (like via a system install).

To avoid that, I think this project should always statically link to GoogleTest.

## Benefits of this change

I work on builds of the RAPIDS libraries (cuDF, cuML, nvForest, etc.) where we do something like the following:

* build all projects and their dependencies as shared libraries
* also build their test executables
* create a container image with the shared libraries + the test executables
* go run containers on a bunch of different configurations (different hardware, environment variables, etc.) to test

In that setting, it's great to have relocatable, statically-linked test binaries because:

* don't need to care about different projects relying on different GoogleTest versions
* don't need to install GoogleTest in the container at all
* don't need to care about the exact location that test binaries are installed

## Reproducible Example

Given tests built like this:

```console
# build tests
$ rm -rf ./build
$ cmake -B build -S . -DBUILD_SHARED_LIBS=ON -DBUILD_CPP_TEST=ON
$ cmake --build build -j8
```

Try moving the test binary to another place and then deleting the build directory.

```console
# relocation 
$ TMP_DIR=$(mktemp -d)
$ mv ./build/treelite_cpp_test "${TMP_DIR}/"
$ rm -rf ./build
$ pushd "${TMP_DIR}"
$ ./treelite_cpp_test
./treelite_cpp_test: error while loading shared libraries: libgtest.so.1.14.0: cannot open shared object file: No such file or directory

$ ldd ./treelite_cpp_test
        linux-vdso.so.1 (0x000076beb360e000)
        libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x000076beb359d000)
        libgtest.so.1.14.0 => not found
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x000076beb2e00000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x000076beb34b4000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x000076beb3486000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x000076beb2a00000)
        /lib64/ld-linux-x86-64.so.2 (0x000076beb3610000)
```

## Notes

I'm planning to propose a PR for this.

Thanks for your time and consideration.

jameslamb:
Closes #704 

Proposes:

* statically linking GoogleTest
* getting GoogleTest via CPM for simplicity and cache-friendliness (mostly borrowing the approach I added to XGBoost in https://github.com/dmlc/xgboost/pull/12494)

### How I tested this

Using the reproducible example from #704.

```console
# build tests
$ rm -rf ./build
$ cmake -B build -S . -DBUILD_SHARED_LIBS=ON -DBUILD_CPP_TEST=ON
$ cmake --build build -j8

# relocate the test executable
$ TMP_DIR=$(mktemp -d)
$ mv ./build/treelite_cpp_test "${TMP_DIR}/"
$ rm -rf ./build
$ pushd "${TMP_DIR}"

# try testing
$ ./treelite_cpp_test
...
[==========] 29 tests from 9 test suites ran. (11 ms total)
[  PASSED  ] 29 tests.

$ ldd ./treelite_cpp_test
        linux-vdso.so.1 (0x0000766f56a2e000)
        libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x0000766f569bd000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x0000766f56200000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000766f568d4000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x0000766f568a6000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000766f55e00000)
        /lib64/ld-linux-x86-64.so.2 (0x0000766f56a30000)
```

### DIFF
--- a/cmake/CPMFindGTest.cmake
+++ b/cmake/CPMFindGTest.cmake
@@ -1,0 +1,65 @@
+# Description:
+#
+#   Find / build GTest. Prefers a system install if one is found,
+#   otherwise downloads sources with CPM (https://github.com/cpm-cmake/CPM.cmake)
+#   and builds GTest.
+#
+# Targets:
+#
+#   GTest::gtest
+#   Gtest::mock
+#
+# Usage:
+#
+#   include(CPMFindGTest.cmake)
+#
+
+function(_treelite_find_gtest gtest_version)
+
+  # Prefer a system-installed GTest if one is found
+  find_package(GTest ${gtest_version})
+  if(GTEST_FOUND)
+    return()
+  endif()
+
+  # not found, use CPM to download sources and build it
+  message(STATUS "GTest not found, fetching GTest via CPM.")
+
+  # For MSVC, ensure GTest uses the same C runtime (/MD vs. /MT) as Treelite.
+  #
+  # As of v1.14.0, googletest did literal string replacements on CMAKE_CXX_FLAGS,
+  # and doesn't respect CMAKE_MSVC_RUNTIME_LIBRARY.
+  #
+  # TODO(jameslamb): remove this once this project uses at least GTest v1.18.0
+  # That version has this fix: https://github.com/google/googletest/pull/4877
+  #
+  set(gtest_force_shared_crt ${FORCE_SHARED_CRT} CACHE BOOL "" FORCE)
+
+  # populate local CPM cache with googletest source (if not already populated), build targets
+  include(${treelite_SOURCE_DIR}/cmake/CPMSetup.cmake)
+
+  # prefer statically linking GTest, so test executables are relocatable
+  CPMAddPackage(
+    NAME googletest
+    GITHUB_REPOSITORY google/googletest
+    GIT_TAG v${gtest_version}
+    VERSION ${gtest_version}
+    EXCLUDE_FROM_ALL YES
+    OPTIONS
+      "BUILD_GMOCK ON"
+      "BUILD_SHARED_LIBS OFF"
+      "CMAKE_POSITION_INDEPENDENT_CODE ON"
+      "INSTALL_GTEST OFF"
+  )
+
+  if(MSVC)
+    foreach(target gtest gmock)
+      # For MSVC, ensure GTest targets propagate the same MSVC C Runtime
+      # to anything linking against them.
+      set_target_properties(${target} PROPERTIES
+          MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}")
+    endforeach()
+  endif()
+endfunction()
+
+_treelite_find_gtest(1.14.0)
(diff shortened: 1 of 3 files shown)


cmake/CPMFindGTest.cmake
```diff
@@ -0,0 +1,65 @@
...
+
+  # not found, use CPM to download sources and build it
+  message(STATUS "GTest not found, fetching GTest via CPM.")
+
+  # For MSVC, ensure GTest uses the same C runtime (/MD vs. /MT) as Treelite.
+  #
+  # As of v1.14.0, googletest did literal string replacements on CMAKE_CXX_FLAGS,
+  # and doesn't respect CMAKE_MSVC_RUNTIME_LIBRARY.
+  #
+  # TODO(jameslamb): remove this once this project uses at least GTest v1.18.0
```
chyunsu3: I'd love to use the latest GTest. Is there a disadvantage for upgrading?

jameslamb: Oh sure! I'm not aware of any disadvantage. Looks like it still has a CMake minimum of 3.16 ([build link](https://github.com/google/googletest/blob/efbef50b594e3f6f3fdeb77b9b4e1ab34e4d1e0f/CMakeLists.txt#L4)), and CI could tell us if it has any compiler support limitations or other problems like that (but I really doubt it for that project).

Also... statically linking it makes it even safer to upgrade this project to a different version 😁 

I'll try that here.

jameslamb: I just tried locally on my Mac and tests seem to work fine with the latest version ([1.18.0, released 3 weeks ago](https://github.com/google/googletest/releases/tag/v1.18.0)).

Pushed https://github.com/dmlc/treelite/pull/705/commits/f91ecf07f5f987bc1b7835f7aeb4903bef7ef188 with that upgrade + removing the MSVC stuff.

chyunsu3 (approved):

chyunsu3:
Thanks!

